PR #81, opened 2015-12-04
Tables

oliver----:
Initial implementation. Not tested.
Some of the things as noted in comments should go into substance core.


oliver----:
Sorry, wrong base branch.

PR #135, opened 2016-02-25
Functions

michael:
First iteration on integrating functions with the backend.

We still don't update the list of available functions automatically. However the improved FunctionComponent now fetches the function spec on demand. It displays a loading state while the spec is loaded.

I put a bunch of todo items to give you some pointers for where to iterate next. 

You may want to study the lifecycle section of FunctionComponent as that's the general pattern to 'synchronize' async data for a synchronous rendering. The component is modelled as a state machine describing transitions. In our case it's just loading (func is null) or loaded (func is set).

### DIFF
--- a/web/sheet/engine/SheetRemoteEngine.js
+++ b/web/sheet/engine/SheetRemoteEngine.js
@@ -1,32 +1,43 @@
 'use strict';
 
-var oo = require('substance/util/oo');
 var RemoteEngine = require('../../RemoteEngine');
 
 function SheetRemoteEngine() {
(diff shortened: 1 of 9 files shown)


nokome:
@michael : thanks for this. Will take a look and if I need to do anything will work on the functions branch

michael:
Feel free to merge this. Just wanted to give you a place to review and check if it breaks things. It's not really complete yet, but hope you can pick it up from there. I don't have much time in the coming days, but if you have questions ping me.

PR #205, opened 2017-02-02
remark expects trailing newlines

sethvincent:
I realized that when remark stringifies the markdown it always has a trailing newline.

I think it makes sense to have the trailing newlines, though I'm not sure if including newlines here will break other tests.


nokome:
Thanks @sethvincent  - this has been an ongoing conundrum for me - to ensure a trailing newline in generated Markdown, HTML etc. There are pros and cons both ways but I tend towards having them. So, I'll accept this now. It'll probably break other tests but that's OK.

PR #219, opened 2017-03-15
Remove sheet

oliver----:


nokome:
@oliver---- : doing a tidy up, could you delete this branch if it is no longer needed

PR #249, opened 2017-03-24
Fix cell code editor

nokome:
@michael : as requested, note that I ended up going with 14px font-size, easy to change if this does not suit.


michael:
Thanks! I will just merge this into my cell-ui branch instead of master. Will create a new PR for it.

Issue #283, opened 2017-04-17
feature-remote-contexts: ability to execute in remote contexts


nokome:
Currently, `run()` and `call()` cells can only execute code in an in-browser `JsContext`. The `feature-remote-contexts` branch adds support for executing cells in R, Python, SQLite and Node execution contexts by:

- [x] registering or discovering "peers" - other Stencila `Hosts` which support the instantiation of `RContext`, `PythonContext` etc
- [x] using the `ContextHttpClient` class to instantiate and connect to a remote context 
- [x] adding these contexts as options in the "language dropdown" (really it should be a context dropdown)

nokome:
Woooooooot! Cells with execution in remote contexts...

![image](https://cloud.githubusercontent.com/assets/1152336/25167467/b5cf8b30-2533-11e7-8676-b722e582b5be.png)

michael:
Wow! Nice work! :)

For the language dropdown.. can you just put some ugly buttons there, and I will try to find some time and rework. Want to develop a Menu abstraction, that we can use for different purposes in Stencila and Texture.

nokome:
Closing this. Plenty of UI stuff still to do (like better selection of context, and stop execution etc). Looks like this for now:

![image](https://cloud.githubusercontent.com/assets/1152336/25561416/f87be2e6-2d1f-11e7-8b5e-dd8ab859c674.png)

PR #285, opened 2017-04-19
Provides for cells with execution in remote contexts

nokome:
This is still a wok in progress, don't merge yet. See #283.

### DIFF
--- a/examples/document/app.js
+++ b/examples/document/app.js
@@ -3,10 +3,11 @@
   using a stub backend.
 */
 
-import { DocumentPage, MemoryBackend, getQueryStringParam } from 'stencila'
+import { DocumentPage, Host, MemoryBackend, getQueryStringParam } from 'stencila'
 
 window.addEventListener('load', () => {
   window.documentPage = DocumentPage.mount({
+    host: new Host(),
     backend: new MemoryBackend(window.GUIDES),
     documentId: getQueryStringParam('documentId') || '01-welcome-to-stencila'
   }, window.document.body)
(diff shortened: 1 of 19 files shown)


nokome:
@michael and @oliver---- : if you get a chance to have a look at this that would be great, otherwise, no probs I can merge it myself.

michael:
The code looks good to me! If I wanted to try it out and run, what setup do i need to for instance execute Python code?

nokome:
@michael : thanks! It would be great if you could user test it: install the python package:

    pip install --user https://github.com/stencila/py/archive/master.zip

and then start the host

    import stencila
    stencila.host.start()

The document should discover the host after a few seconds and allow you to use `py` or `sqlite` context. 

Oh my, so much to do for next week!

michael:
Which Python version? I have 2.7.10 (preinstalled on OSX) but there's no pip package manger installed.

Then i run the python script as you described..  and for the Stencila UI, do I need to run the desktop app or does discovery work via the development environment?

nokome:
2.7 is fine. pip is easiest way to install from GitHub.

Discovery works for dev setup.

By the way, you can test external code via the stencila/node package too by
running `npm start`.




On Apr 26, 2017 9:22 PM, "Michael Aufreiter" <notifications@github.com>
wrote:

Which Python version? I have 2.7.10 (preinstalled on OSX) but there's no
pip package manger installed.

Then i run the python script as you described.. and for the Stencila UI, do
I need to run the desktop app or does discovery work via the development
environment?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/stencila/stencila/pull/285#issuecomment-297365669>, or mute
the thread

PR #291, opened 2017-05-07
WIP: Add JupyterContextClient

nokome:
The `JupyterContextClient` aims to allow users to execute cells within Jupyter kernels.


nokome:
Hi @rgbkrk and @michael : pinging you here just to let you know that I've started this integration with Jupyter. Instead of implementing a relay, the current approach is to use the Jupyter JS source for client-kernel communications as a vendor bundle. He aim is to have `JupyterContextClient` simply translate calls to the Stencila Context API to messages in the Jupyter API and use the Jupyter code base for doing the actual comms between the browser and the kernel. Not sure it's the right approach but it seemed the most expedient at this time.

There is also a WIP pull request at https://github.com/stencila/node/pull/15 which will launch Jupyter kernels.

rgbkrk:
Awesome!

There's a module available for launching kernels here:

* https://github.com/nteract/spawnteract

That's currently in use on Sage Math Cloud, nteract, and Hydrogen.

To find out what kernels are available, there's `kernelspecs`:

* https://github.com/nteract/kernelspecs

I'm looking forward to this work! I'll be on vacation this week but may end up perusing issues and fixing a few things when there's downtime.

nokome:
@rgbkrk : Thanks heaps for pointing out `spawnteract` and `kernelspecs` - super useful!

Can you recommend anyone I could work with on the client side JS while I'm here in the Bay Area? I'm hopeful someone with experience with that code base could point me in the right direction and save a lot of work.

nokome:
OK so my poor understanding of networking has been exposed! Thanks to @Carreau, Mike Pacer and @fperez for help with this.

I was under the impression that the Jupyter notebook client was using ZeroMQ over Websocket. Instead, it is using Websocket to a relay server in Python that then passes on the messages using ZeroMQ to the kernel. In other words, the client is not talking directly to the kernel - which was the architecture I envisaged that we would use here.

There are proposals and tools for ZeroMQ over Websocket but currently they require ZeroMQ extensions which presumably are not supported by existing Jupyter kernels: https://rfc.zeromq.org/spec:39/ZWS/ .https://github.com/zeromq/JSMQ

So, we need to implement the HTTP/Websocket-ZeroMQ relay approach. As @rgbkrk previously pointed out https://github.com/nteract/kernel-relay is a prototype of this in Node.js. Over in https://github.com/stencila/node/pull/15 we should do something similar to `kernel-relay` by using https://github.com/nteract/enchannel-zmq-backend but with our existing HTTP API.

Then, here in the client, instead of using the old JS code from [/jupyter/notebook](https://github.com/jupyter/notebook/), we should use [Jupyter Lab Services package](https://github.com/jupyterlab/jupyterlab/tree/master/packages/services). Alternatively, it may be that we can do the just use the existing `ContextHttpClient` to talk to the relay.

I'll work on the Node.js side first and then come back here. Aiming for this to be ready for the 0.27 release at the end of this month.

rgbkrk:
The other alternative is to use `zeromq` directly, that's what we do in nteract with the `jmp` module. The messages are the same.

nokome:
@rgbkrk : thanks for pointing out `jmp` . But I think it is best to stick with a relay approach that you first suggested so that the front-end can live in a web browser i.e for when documents or sheets are published as HTML pages

nokome:
Closing this in favour of stencila/node#15.

Issue #295, opened 2017-07-03
Cell errors are not displayed when a variable is assigned


nokome:
For example:

![image](https://user-images.githubusercontent.com/1152336/27800836-f8008242-6012-11e7-918c-88bd3c6002e3.png)

versus:

![image](https://user-images.githubusercontent.com/1152336/27800813-d9937080-6012-11e7-992d-bf26d2d5c8a8.png)

this does not happen with Mini cells:

![image](https://user-images.githubusercontent.com/1152336/27800883-35f7b660-6013-11e7-839b-d0487b7c65a5.png)

oliver----:
Fixed.

PR #304, opened 2017-07-05
Propagate errors

oliver----:
- Fixing #295
- Getting CI running again


oliver----:
only one test is failing now.

PR #320, opened 2017-07-16
Allows for more env vars in dev; adds STENCILA_DISCOVER

nokome:
Allows closer mimicking of Desktop deployment during development


michael:
This is looking good 👍

PR #324, opened 2017-07-22
WIP: Datatable schema

nokome:
A schema for Datatables. Tries to avoid too much wheel reinventing by building upon the Fictionless Data specifications as much as possible. Currently preliminary, expect significant changes.

### DIFF
--- a/src/datatable/DatatableDataModel.md
+++ b/src/datatable/DatatableDataModel.md
@@ -1,0 +1,45 @@
+# Data model for a Datatable
+
+- a requirements/specifications document so that we can decide on the best approach to developing a XML schema for Datatables
+
+- this document will either need to be updated to be consistent with the developed schema, or deleted
+
+- initially, for determining approach, we should focus on the MUSTs and worry less about the OPTIONALs
+
+- a Datatable is similar to database table, it has one or more **columns** (a.k.a _fields_) and for each column there is a value in each **row**
+
+- although there is an emphasis on a column wise schema for Datatables this does not necessarily mean that the in-memory representation needs to be column-wise 
+
+- a Datatable can have OPTIONAL meta data such as **name**, **title**, **description** and **authors** which the user SHOULD be able to add and modify
+
+- Each **column**,
+
+    - MUST have a **name**, this name SHOULD be able to be set and changed by the user, and/or be defined by the external data source (e.g. a CSV file with a header)
+
+    - MUST have a **type** which specifies the _type_ of data that can be in that column:
+
+        - the user SHOULD be able to change the type of a column
+
+        - types might include string, number, integer, boolean, date
+
+        - the default type is "string"
+
+        - having types allows for editing restrictions on the type of data that a user can enter into a cell
+
+        - having types allows for optimisations in storage and analysis across the Datatable
+
+    - can have one or more OPTIONAL **constraints** which specifies the acceptable _values_ in each column
+
+    - can have an OPTIONAL **title** string which the user SHOULD be able to modify
+
+    - can have an OPTIONAL **description** string which the user SHOULD be able to modify
+
+- Similarity and difference to `Sheets`:
+
+  - Superficially the UI for `Datatables` is similar to that for `Sheets` in that they are both grid based
+
+  - However, the _data model_ is quite different. Sheets can be though of as a graph of cells, each specifying their dependency on other cells via cell formulae. In a spreadsheet those cells are displayed in a grid. But that is really a UI convention, not a reflection of the underlying data model. The position of a Sheet cell in that sheet does not determine the data content of that cell. A cell in column A1 is not restricted to having the same type of value as a cell in A2 even though they are in the same column. In contrast, in a Datatable, cells in a column are restricted to having the same type.
+
+  - A Sheet tends to be more sparse having more empty cells. It may be best to model sparse Sheets as a graph of cells having a particular position e.g. `A1` rather than them being either column or row oriented data
+
+  - Where sheets are dense, and thus more suitable to row or column based operations, that is usually because they contain a lot of data - in those cases we would recommend that the user use a Datatable for that data
--- a/src/datatable/DatatableSchema.rng
+++ b/src/datatable/DatatableSchema.rng
@@ -13,7 +13,6 @@
         Based on the Tabular Data Resource specification (https://specs.frictionlessdata.io/tabular-data-resource/) but with modifications to suit a DatatableEditor.
       </a:documentation>
       <ref name="datatable-meta"/>
-      <ref name="datatable-schema"/>
       <ref name="datatable-fields"/>
     </element>
   </start>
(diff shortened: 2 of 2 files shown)


oliver----:
> Re the proposed custom schema in https://github.com/stencila/stencila/pull/324. Let's move the discussion regarding that over there so we can get more specific about the bits that we like and don't like. I'm not sure what you mean by the "RDF thing" (I know RDF is but I can't see where it comes into that schema). I agree that the relational aspect of that schema (`<element name="primaryKey">` and `<element name="foreignKeys">`) should be dropped. I put them in for completeness/consistency with the Frictionless Table Schema spec as the starting point for these discussions but they go beyond the simple tabular data editor. 

Yep, that's what I meant.

oliver----:
> I don't agree that  we should treat `Datatables` as `Sheets` "with restrictions" - at least not to the extent of having a shared XML schema. I think they are distinct enough to warrant separate schemas and I'm concerned that trying to tie them together in a shared schema would constrain both.

nokome:
@oliver---- : OK, thanks. I've removed that stuff. 

Should we progress by modifying this schema, or do you think it's better to start from scratch with something very minimal and build it up?

oliver----:
> I don't agree that  we should treat `Datatables` as `Sheets` "with restrictions" - at least not to the extent of having a shared XML schema. I think they are distinct enough to warrant separate schemas and I'm concerned that trying to tie them together in a shared schema would constrain both.

Ok. So before we define a schema, let's put together requirements, and identify the differences. I had the feeling that both are pretty similar. My concern is that if we go into a totally different direction, that we won't be able to share anything between sheets and data tables. E.g., if data tables store values in columns, and sheets in rows, it is impossible to share the same insert row / insert column logic.

oliver----:
OTOH, examples of how I see data table as a restriction:
- there are no formulas in cells, and no formatting etc. -- cells are very pure.
- column type is strict, as opposed to sheets, where it could be the default 'formatting', but in sheets you can override formatting of a single cell

oliver----:
> Should we progress by modifying this schema, or do you think it's better to start from scratch with something very minimal and build it up?

Let's agree first on a basic direction.

nokome:
I've added a new (temporary?) document describing data model requirements for Datatables and similarities/differences with Sheets:

https://github.com/stencila/stencila/blob/datatable-schema/src/datatable/DatatableDataModel.md

@oliver---- feel free to add to that, or comment here.

I'm willing to be convinced otherwise, but at present I'm still of the view that in the long run, using the same _data model_ for Sheets and Datatables would be more work and produce inferior products, and outweigh any benefits that might come from of shared data model implementations. Instead perhaps we should focus on potential efficiency gains from shared UI implementations. Acknowledging, of course, that you and @michael are best placed to judge that :smile:

oliver----:
@nokome Thanks. Let's continue refining that.

I don't think, that sparsity is a good criteria to distinguish Sheets and Datatable.
I thought the key difference is dynamic vs static cell content. The sheets I have created are typically rather dense, having only some header lines to describe blocks, and adding some dynamic columns, or only some aggregating cells at the bottom of a block. Probably it depends on the use-case how much is evaluated dynamically, and how much is static data. E.g. a CSV file plus an aggregated column could be very similar to a Datatable for the user. At which point should they switch to a Sheet?
So, I am not yet convinced that we need two fundamentally different things. 

I admit, I see a great value of a Datasource Management tool allowing to manage the schema together with the data, with means for aggregating and cleaning data etc. But that is a challenge on its own, and probably should be a different application. I don't think it is just a side-quest of Sheets ;)

oliver----:
Just to complement the idea of transitioning from Datatable to Sheet -- and because I just watched a talk about Typescript, which made me think into this direction: what about opt-in typing in Sheets.
In a Sheet, typing is dynamic by default, i.e. it does auto-detecting to derive the formatting. Still you can select the whole column and change the type. In Excel et al. this is just interpreted as formatting directive. However we could also look at it as static typing opt-in. Could improve the tooling in Mini as well, if we knew the type in advance.

src/datatable/DatatableDataModel.md
```diff
@@ -0,0 +1,45 @@
...
+
+    - can have an OPTIONAL **title** string which the user SHOULD be able to modify
+
+    - can have an OPTIONAL **description** string which the user SHOULD be able to modify
+
+- Similarity and difference to `Sheets`:
+
+  - Superficially the UI for `Datatables` is similar to that for `Sheets` in that they are both grid based
+
+  - However, the _data model_ is quite different. Sheets can be though of as a graph of cells, each specifying their dependency on other cells via cell formulae. In a spreadsheet those cells are displayed in a grid. But that is really a UI convention, not a reflection of the underlying data model. The position of a Sheet cell in that sheet does not determine the data content of that cell. A cell in column A1 is not restricted to having the same type of value as a cell in A2 even though they are in the same column. In contrast, in a Datatable, cells in a column are restricted to having the same type.
```
oliver----: I think the key of success of the Sheet paradigm was the ability to mix data and code, with focus on the data side (as opposed to the flow/dependency part). It is not just a convention. It is a projection of the multi-dimension nature of a program: code, values, and dependencies, into one view.
In imperative programming you see the local dependencies, or order, but need a debugger to reveal the values.
In graphical programming you see the data-flow, plus a symbolic version of the code, not the values.

nokome:
> BTW, I admit, I see a great value in a Datasource Management tool allowing to manage the schema together with the data, with means for aggregating and cleaning data etc. But that is a challenge on its own, and probably should be a different application. I don't think it is just a side-quest of Sheets ;)

Yes, that is what I thought we had agreed that Datatables would be - a different application to Sheets. We discussed doing more advanced data cleaning ala OpenRefine and decided that would be too big a job right now and so scaled back to the original idea just having a simple tabular data editor that was separate from Sheets. Something similar to CommaChalemeon which can be deployed in the Desktop and, although it is very simple, seems to be useful for people (based on what the Open Knowledge Foundation have told us). 

It feels like it is overcomplicating things, at least for now, to try and find a ways to merge the two use cases.

oliver----:
Ok, now we have found the source of confusion 😄 
You are envisioning Datatables being something like Microsoft Access (plus OpenRefine etc.). And we were on a much more light-weight track.
We believe that for dealing with tabular data, such as CSV, a Sheet with column meta and types is sufficient. I don't see that we have enough resources to pursue Datatables, because it is substantially so different.

nokome:
> Ok, now we have found the source of confusion :smile:

No, I don't think there was any confusion in the first place. I think we were just discussing pros and cons of alternative data models.

> You are envisioning Datatables being something like Microsoft Access (plus OpenRefine etc.). And we were on a much more light-weight track.

No, I envision Datatables being a simple light weight tabular data editor just like you (we got into discussions in Linz about doing more complicated stuff but all agreed we didn't have the resources to go there right now).

> We believe that for dealing with tabular data, such as CSV, a Sheet with column meta and types is sufficient. 

Yes, I agree with "column meta and types" and that is what is reflected in the `DatatableSchema.rng` and `DatatableDataModel.md`. I think the disagreement is just in the "a Sheet with" - I'm leaning towards a separate data model for Sheets and Datatables, you're leaning towards a combined data model.

> But with typed Sheets I think we could get pretty far, and then we have the synergies.

To reiterate, my concern is that because Sheets and Datatables are fundamentally different that we will unnecessary constrain both products by trying to use the same data model. I question whether the synergies are at the UI level, rather than the data model level.

But, since we agree on what we are trying to achieve, and since you know far more than me in terms of what is going to be needed to implement this, **I'm happy to defer to your opinion on this. I suggest that the most productive way forward is for you to go ahead with a "typed Sheets" approach.** If that doesn't constrain either product and delivers efficiencies then that would be awesome, and if not we can always refactor later.

nokome:
Just gotta say that the more I think about this...

> In a Sheet, typing is dynamic by default, i.e. it does auto-detecting to derive the formatting. Still you can select the whole column and change the type. In Excel et al. this is just interpreted as a formatting directive. However we could also look at it as static typing opt-in. 

(i.e. allowing the user to statically type a cell, range of cells or an entire column (and also name a column))

and this...

> a Sheet with column meta and types is sufficient

(i.e. making a Datatable a special Sheet in which all columns are named and typed)

the more I like it and the more I see the synergies. Once again, you are right @oliver---- .

oliver----:
@nokome I like that it is so hard to convince you 😉 , because you have good reasons. Let's see how far we get with this approach.

PR #333, opened 2017-08-30
Create 2017-Roadtrip-Roadmap.md

daniellecrobinson:
What do you think of this format for a user facing road trip roadmap?


nokome:
I like the format (including the emoji :smile:) (obviously the timing of what functionality will be available when needs to be more realistic, but I understand this is just about the format :smile: ). Let's move this over to a GDoc and then put it up in the community forum.

Issue #310, opened 2017-07-06
Introduce definition of external functions


oliver----:
We want to be able to expose a new function implemented in a cell something like
```
myfun = function(x:int, y:int)
return x+y
```

nokome:
I think that this should be a fairly high priority so that we can define built in Mini functions using this approach.

To be be able to specify the external language, and to specify the type of the return value, the syntax needs to be something like this (which is very similar to Typescript's typing syntax):
```
myfunc = function js(x:int, y:int):int
return x+y
```

Do we need to allow type specification to be optional? What are the pros/cons of making it optional or not? e.g.
```
myfunc = function js(x, y)
return x+y
```

Multiple dispatch will be very useful for keeping a tidy function namespace in Mini and should be implemented as part of this. See [src/js-context/functions/types/_multifunc.js](src/js-context/functions/types/_multifunc.js) for our current implementation of that.

nokome:
Thinking some more about how Mini function definitions would actually be defined...

A user defined R function like this

```
myfunc = function r(x:int, y:int):int
return x + y
```

Could be implemented as a Mini function in Javascript as:

```js
const myfunc = {
  name: 'myfunc',
  inputs: [
    {
      name: 'x',
      type: 'int'
    },{
      name: 'y',
      type: 'int'
    },
  ],
  output: {
    type: 'int'
  },
  language: 'r',
  code: 'return x + y'
}
```

A builtin Mini function with some additional documentation could look something like:

```js
const sum = {
  name: 'sum',
  description: 'Adds up some numbers',
  inputs: [
    {
      name: 'values',
      type: '[number]',
      description: 'The numbers to add up'
    }
  ],
  output: {
    type: 'number'
  },
  language: 'js',
  code: function(values) { return stdlib.sum(values) }
}
```

nokome:
Another proposal for moving ahead with this...

We should treat functions just like other parts of the platform and make them "first class citizens" just like `Documents` and `Sheets`. We would define a `Function` data model using an XML schema (as we are doing for `Sheets`, and will do for `Documents`) and UI components that would allow editing of functions as standalone entities along the lines of what @michael proposed [here](https://medium.com/@_mql/reproducible-science-for-people-who-dont-code-5850abe806db).

In the short term, I don't think we should do any UI for `Functions` and instead just focus on:

- defining a data model
- wrapping existing Javascript functions (e.g. from stdlib) to quickly establish functionality in Mini (although we should be careful not to rush naming!). 
[...]

PR #336, opened 2017-09-08
WIP: Functions

nokome:
Work in progress to add `Functions` as "first class citizens" just like `Documents` and `Sheets`. See #310 

This requires the `substance#improvements` branch.

Try `node make test:one -f "tests/function/functions.test.js"`


michael:
Interesting to provide multiple implementations (different languages) for one function. I wonder how we choose one implementation on execution?

nokome:
I envisage three criteria for choosing amongst implementations:

1. Availability: if you don't have an execution context capable of running an implementation (e.g. in R), then you'd have to use on of the other implementations (e.g. JS)

2. Proximity: where possible execution should be performed in the same context as the data, rather than transferring data between contexts e.g if you have a 1mil row data frame in an R context then it's better to execute `aggregate` there instead of transporting the data to the clients to run in JS

3. Efficiency: faster implementations should be preferred over slower ones. Although in most instances proximity is likely to outweigh computation speed, in some instance it may make sense to transfer whole parts of the execution graph to another, faster context.

michael:
That sounds like a smart strategy! :) Let's just try not to optimize from the start. I agree the format should allow for multiple implementations. However we could start off with always selecting the first lang, so we can push implementing that selection logic to later (could be a source of errors too).

nokome:
Yeah, totally. Just trying to get the schema to be flexible enough to allow for these types of things later, like you said "let’s try to start with something we can stick with for a while". Initially, I imagine we'll be focussing on just the one JS implementation.

nokome:
Almost completed a first iteration of this. Still to do:

- XML-based (instead of JSON based) value specifications for default parameter values, and test argument and result values

- check that implementation types are consistent with parameter types when function is defined

- compile implementation code from external files e.g. `sum.fun.xml, sum.fun.js, sum.fun.r`

- quick documentation write up

michael:
Looking good! Could you base this on the 'bloody-edge' branch? We created this one to be able to merge our experimental stuff more frequently.

nokome:
This is now rebased to `bloody-edge` branch. Add `<code include="...">` to allow for source code in external files for each implementation. In the process of simplifying the schema to make definitions less verbose. Then should be close to finalising this first round!

michael:
I'll close this PR and create a new one against bloody-edge, which you can merge once you are ready with the final iteration!

PR #351, opened 2017-09-20
Functions

michael:


src/function/FunctionDocument.js
```diff
@@ -0,0 +1,58 @@
...
+  getDocTypeParams() {
+    return FunctionSchema.getDocTypeParams()
+  }
+
+  getXMLSchema() {
+    return FunctionSchema
+  }
+
+  getRoot() {
+    return this.get('function-1')
```
oliver----: I want to allow to control ids for certain elements. So this will be better in future.

src/function/FunctionSchema.rng
```diff
@@ -15,7 +15,8 @@
       <ref name="function:description"/>
       <ref name="function:params"/>
       <ref name="function:return"/>
-      <ref name="function:source"/>
+      <ref name="function:implems"/>
```
oliver----: Maybe use the full name `<implementations>` and `<implementation>`?

src/function/FunctionDocument.js
```diff
@@ -19,30 +19,70 @@ export default class FunctionDocument extends XMLDocument {
...
+    const $implems = this.getRoot().findAll(`implem[language=${language}]`)
+    let promises = []
+    let lookup = {}
+    for (let $implem of $implems) {
+      let $types = $implem.findAll('types type')
+      let implemName = name + $types.map($type => '_' + $type.text())
+      let code = $implem.find('code').text()
+      // Define the function
+      let promise = context.defineFunction(implemName, code).then(() => {
+        // TODO Use type hierarchy to map between a possible function call signature
```
oliver----: 👍

oliver----:
I think we can move on.
Will rebase this, evaluate, and do the merge

oliver----:
I merge. Will integrate it with my Engine stuff then see how it is going.

PR #352, opened 2017-09-20
Sheets and Docs and Functions

michael:


src/engine/Engine.js
```diff
@@ -0,0 +1,381 @@
...
+      names taken from the signature. The context is used to
+      execute the sourceCode, using the arguments object.
+    - chunk: like with external cells, arguments are provided
+      as object. The source code is run in the same way as we know
+      it from notebook cells, such as in Jupyter.
+  */
+  callFunction(funcNode) {
+    const functionName = funcNode.name
+    // ATTENTION: we removed support for 'external cells' (mini + external source code)
+    // as we want to evaluate how far we can get with just
```
nokome: Worried face :worried:

oliver----: Don't worry too much. If it turns out to be necessary later we bring this back again. 
But this is actually a consequence of your proposal of making external code stronger:better reflection + implicit return :)
Having this makes 'external cells' which are essentially self calling functions pretty obsolete. And having so many concepts for cells/expressions is worrying me personally.

nokome: Removing this is really worrying me! It's a major departure in terms of the functionality of the entire platform. I'm really surprised.

nokome: OK, so now I understand this much better after talking to @oliver---- and think it's a great idea to explore - getting rid of the mini special header functions (e.g. `r()`) and instead doing code dependency analysis - looking forward to it!

PR #353, opened 2017-09-21
Clean up

nokome:
@michael and @oliver---- : since we are in cleanup mode I though I would address a few of the things I have been wanting to cleanup for a while as well as remove some of the dependencies related to the converters. The individual commits are relatively standalone and should be superficial - either garbage removal, renaming or moving around. If you happy with this could you merge as soon as possible, thanks!


michael:
Thanks for this!

PR #354, opened 2017-09-22
WIP: Updates to Functions

nokome:
Some improvement to `Functions` and relevant parts of `JsContext`...

- remove `option.pack` from `JsContext` methods as it is an optimisation that we identified in July as causing confusion.

- fixes and improvements for multi-language dispatch of calls

- Add descriptive elements to the schema based on [our initial work on defining functions](https://github.com/stencila/stencila/issues/310#issuecomment-326141685) which was in turn informed by things like [JsDoc](http://usejsdoc.org/) and [R function documentation](https://cran.r-project.org/doc/manuals/R-exts.html#Rd-format);

  - `<summary>` : short one sentence summary suitable for use in tooltips etc (like JsDoc `@summary`)
  - `<relateds>` : names of related functions (like JsDoc `@see`)
  - `<examples>` : examples of usage (like JsDoc `@example`)

@oliver---- : I am hoping that this will be the end of the first iteration of `Functions`. Could you review and merge. 

I will work on defining functions in https://github.com/stencila/lib and use lessons learned during that for a possible second iteration in the next 2-3 weeks.

### DIFF
--- a/index.es.js
+++ b/index.es.js
@@ -18,3 +18,4 @@ export { address, value }
 export { default as getQueryStringParam } from './src/util/getQueryStringParam'
 
 export * from './src/sheet/index.js'
+export * from './src/function/index.js'
--- a/src/contexts/JsContext.js
+++ b/src/contexts/JsContext.js
@@ -35,9 +35,7 @@ export default class JsContext extends Context {
    *
    * @override
    */
-  runCode (code = '', options = {}) {
-    const pack = (options.pack !== false)
-
+  runCode (code = '') {
     // Create a function from the code and execute it
     let error = null
     try {
@@ -61,7 +59,7 @@ export default class JsContext extends Context {
     }
 
     return Promise.resolve(
-      this._result(error, output, pack)
+      this._result(error, output)
     )
   }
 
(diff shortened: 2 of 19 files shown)


nokome:
> Maybe we need to use CDATA? could there be more cases like this?

Yes maybe. I have avoided it here (because it is so ugly :)) and am happy to use `array[string]` instead of `array<string>`. But also fine to use CDATA if you think that would be better . It will be needed for implementation code which uses the `<` symbol I suppose?

michael:
@oliver----  ok to merge these latest additions and then go from there?

oliver----:
Yeah. Let's move on.

oliver----:
> Yes maybe. I have avoided it here (because it is so ugly :)) and am happy to use array[string] instead of array<string>. But also fine to use CDATA if you think that would be better . It will be needed for implementation code which uses the < symbol I suppose?

Correct.

michael:
Merging this. I would suggest that for core functions we always use external files and include them in a bundle step. For user functions we can store everything in the XML right away.

nokome:
> I would suggest that for core functions we always use external files and include them in a bundle step

I agree. The more I think about it, the more I think the core functions should just be developed as a package of functions using a strict naming convention (for correct dispatching) and built and tested using native tooling for each language (eg roll-up and tape for JS, devtools and test that for R). That makes dev and testing a lot easier and more welcoming.

The XML files will just provide the descriptors. The source code does not need to be injected into it. Instead the package of functions is just imported by `JSContext`, `RContext` etc in the standard way for the corresponding language.

PR #355, opened 2017-09-27
FunctionManager

michael:


oliver----:
@michael I am working on a stubbed integration with the Engine and see if we need to change things.

michael:
First iteration done. Merging into bloody-edge. Texture and Substance deps now point to npm release, mini-core still needs npm link.

PR #360, opened 2017-10-02
Initial version of FunctionUsageComponent

michael:
image

Hacked the EditLinkTool to start prototyping the FunctionUsageComponent. We need to connect this now with the Mini editor, by parsing and analysing the mini expression.


michael:
Merging into bloody-edge for further integration work.

PR #361, opened 2017-10-05
WIP: Cell actions

michael:
We want to be able to change cell setting via keyboard. This will be relevant for spreadsheets as well, as we need a way to switch the language of a cell.

This requires new Substance API's that @oliver---- is working on. 

Don't merge!


oliver----:
I think this is obsolete.

oliver----:
@michael do you still need this branch?

PR #362, opened 2017-10-06
New Context/JsContext API

nokome:
This is a refactoring of the `Context` API with initial implementation for `JsContext`. The methods `runCode` and `callCode` have been replaced by `executeCode` which uses static analysis of the code to determine the inputs (names of variables that his block of code is dependant upon) and output (name of variable that is returned from the code). 

The value returned from :
```
executeCode('let b = 2 * a', {'a' : { type: 'integer', data: 2 } })
```
is:
```
{
    inputs: ['a'],
    output: 'b',
    value: { type: 'integer', data: 4 },
    errors: []
}
```

For `JsContext` the `output` is non-null (i.e. a string) if the last statement in the code is a variable declaration (e.g. `let x = 4`) or an identifier (e.g. `x`). See `tests/contexts/JsContext.test.js` for more examples. This convention will be used for other language contexts.

Note a change in the format of the "packed" value e.g. `{ type: 'integer', data: 2 }` instead of `{ type: 'integer', format: 'text', content: '2' }` (this is part other changes to value packing/unpacking to reduce confusion, make it more efficient and allow for "pointers" to values residing in remote contexts).

This requires `npm link stencila-libcore` at the `js-branch` (which adds a missing export)


oliver----:
Let's call it a first iteration.

PR #426, opened 2017-10-24
WIP: Working on Range expresssions

oliver----:


oliver----:
image

oliver----:
@nokome ^^
Getting closer, but...
having some troubles getting types of composites propagated through the system correctly/consistently.
Need to speed-down on this and do proper architectural design. We should specify this together

nokome:
@oliver: great, thanks! Yes, lets discuss some more

oliver----:
We want to produce `table` values for 2D ranges. @nokome it would be good to have libcore functions (such as `sum()`) support `table` values

oliver----:
stencila-range-expressions

oliver----:
@nokome ^^

PR #458, opened 2017-12-08
Stencila Enhancement Proposals

nokome:
As discussed.

@oliver--- : this just formalises your proposal a little (and puts everything into one folder); of course, feel free to make suggestions/changes

@daniellecrobinson: if you could take a look at fleshing out step-0000.md a little that would be great

We need to avoid making this too burdensome but at the same time it would be great if these StEP documents evolve over time into useful pieces of documentation like PEPs.


oliver----:
Cool. Like it.

nokome:
> Cool. Like it.

That's a relief. Really. I need to stop worrying so much about what you guys think of my PRs :)

oliver----:
You should stop worrying because you are pretty close to perfect.... in way to many regards ;)

oliver----:
Seriously, my main goal is to have a process. Your proposal goes a bit further than I  considered in the beginning. but this is a better suggestion than my try.
You started the guide-line, ... I will try to consolidate the next weeks.
Probably it will be important to see how it works internally. And eventually how it is received by non-tech-savvies.

oliver----:
Let's eat our own dog-food!

oliver----:
Shall we merge?

nokome:
OK, let's merge.

> Probably it will be important to see how it works internally. 

Yes absolutely, needs to work for the devs foremost.

> And eventually how it is received by non-tech-savvies.

I don't think this should ever try to reach an audience other than developers - these are intended to be *technical* design and spec docs - user guides and community forum are there for other audiences

PR #462, opened 2017-12-08
CellEditor - Iteration I

oliver----:


oliver----:
Shall I review?

michael:
Too early I would say... just tinkering around to get aware of all pieces involved, and set up new architecture and workflow.

oliver----:
@michael The slow response after confirming a cell, was related to the hierarchical listener. I have disabled this in substance. Let's see when we really need this, and then address it in the right way.

oliver----:
We should do some tidy-up and then get this merged.

src/sheet/SheetEditor.js
```diff
@@ -350,8 +350,25 @@ export default class SheetEditor extends AbstractEditor {
...
     let cell = this._getAnchorCell()
     let newValue = this._formulaEditorContext.node.getText()
+    let { anchorRow, anchorCol} = this._getSheetSelection()
+
+    // Select cell below
+    anchorRow += 1
+
     editorSession.transaction(tx => {
       tx.set(cell.getPath(), newValue)
+      tx.setSelection({
```
oliver----: There is a method in SheetComponent, which also shifts the viewport:
```
this.refs.sheet._nav(1, 0)
```
On the long run we should tidy-up the API of SheetComponent

oliver----:
@michael it would be good to rebase onto develop, resolving the merge conflict

michael:
I'm doing that right now.

src/sheet/SheetEditor.js
michael: Wanted to do it one flow. I guess we should have the _nav as a utility returning a new selection, and handle the this.send stuff in a separate method.

oliver----: I see. Yeah, I need to think how to make the SheetComponent API useful.
So let's stick with your implementation, and fix SheetComponent in a different iteration

src/sheet/SheetEditor.js
```diff
@@ -350,25 +350,10 @@ export default class SheetEditor extends AbstractEditor {
...
-          type: 'range',
-          anchorRow,
-          anchorCol,
-          focusRow: anchorRow,
-          focusCol: anchorCol
-        },
-        surfaceId: this.refs.sheet.getId()
-      })
+      let newSel = this.refs.sheet.shiftSelection(1, 0, false)
+      tx.setSelection(newSel)
```
michael: @oliver---- better like that?

oliver----: Yeah. Better that we don't see the internal selection data.
And with re viewport: SheetComponent should be smart enough to update the viewport automatically
If it doesn't we will fix it later.

oliver----:
I think it is good for this iteration.

oliver----:
Sweet.

oliver----:
Let's call it a PR. Want to push da button?

oliver----:
Ersta :)

michael:
@oliver---- I don't understand how this code works 🤷‍♂️

https://github.com/stencila/stencila/blob/a38239ab900a9a17cde36b69e76826304e2309c7/src/sheet/SheetComponent.js#L383-L399

michael:
like an anonymous block thing...

oliver----:
Jui. Yeah that's ugly. Probably I have moved this from somewhere else and 'fixed' it using the block.
But honestly, a bit embarrassing ;) need to fix. Thx for the TODO

> Note: it is working because newFocusRow/Col has been determined above that block.

oliver----:
@michael This was good work! Thx

PR #464, opened 2017-12-11
Upgrade libcore and mini

nokome:
These upgrades introduce operator functions e.g. `add`, `multiply` for Mini. Operators are treated just like any other functions and handled within libcore rather than by `MiniContext`


nokome:
@oliver---- can you please review and merge if acceptable

nokome:
Also, this version of libcore includes a `sleep(seconds)` function for testing UX of cell editor when long running calls are involved.

src/sheet/SheetEngineAdapter.js
```diff
@@ -181,11 +181,9 @@ class CellAdapter {
       this._lang = null
       let sheet = cell.getDocument()
       let type = sheet.getCellType(cell)
+      let text = cell.text()
       // HACK: this marshalling should be done somewhere else
```
oliver----: Seems that you fixed this HACK

oliver----: If so, you could remove this comment

oliver----:
A little weird but consequent:
image

src/sheet/SheetEngineAdapter.js
nokome: I'm not sure. This was quite a quick fix and still feels temporary. I'm going to propose a change in how we deal with types, including casting, so we may deal with this comment then.

oliver----: ok

oliver----:
stencila-operators-are-functions

👍  Working for me

nokome:
> A little weird but consequent

Yes, I suggest that we deal with this in the `FunctionUsageComponent` - don't display for operator functions.

PR #475, opened 2017-12-17
Function helper context.

Integral:


src/shared/FunctionReferenceComponent.js
```diff
@@ -15,12 +15,65 @@ export default class FunctionReferenceComponent extends Component {
...
+            $$('span').addClass('se-description').append(param.description)
+          )
+        )
+      })
+
     } else {
       const functionList = functionManager.getFunctionNames()
+      functionList.forEach(func => {
+        el.append(
+          $$('div').addCLass('se-item').append(func)
```
oliver----: typo: addClass

oliver----:
In a next iteration we should enable this feature in Document as well.

michael:
Let's make this feature a responsibility of ProjectComponent, then we have it in Document and other types (e.g. Notebook) to come. :)

PR #476, opened 2017-12-17
Cell referencing

Integral:
Not ready yet.

### DIFF
--- a/src/sheet/SheetEditor.js
+++ b/src/sheet/SheetEditor.js
@@ -10,6 +10,7 @@ import FormulaEditor from './FormulaEditor'
 import FunctionUsageCommand from '../shared/FunctionUsageCommand'
 import FunctionUsageTool from '../shared/FunctionUsageTool'
 import CodeEditorPackage from '../shared/CodeEditorPackage'
+import { getCellLabel } from './sheetHelpers'
 
 export default class SheetEditor extends AbstractEditor {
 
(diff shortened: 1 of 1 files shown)


oliver----:
OMG :)

oliver----:
image

oliver----:
dxdnwllbeq

oliver----:
@nokome ^^

nokome:
Woah, awesome @Integral !

michael:
Yeah, bravo! :)

Next we need a addCellMarkers API to color the range while dragging but also other ranges in the expression. What is also important is a tokenizer so we can implement a replace range functionality as well as detecting all the ranges of an expression.

oliver----:
The tokens are there. Just need to provide an API to get the current token.

Integral:
It was Michael's idea, I'm just an implementer behind ;)

michael:
Cool. We thought of adding such API's on a custom CellEditorSession. The CodeEditor would be the wrong place as there are multiple instances of it?

oliver----:
The tokens are on the cell.

michael:
Thought of something like this:

```js
cellEditorSession.replaceSelectedToken('A1:C5')
```

or better just a helper?

```js
replaceSelectedToken(cellEditorSession, 'A1:C4')
```

michael:
Merging this and calling it a day. See you tomorrow!

oliver----:
Cool. CU 2morrow.

PR #479, opened 2017-12-19
Cell referencing.

Integral:


oliver----:
Good enough for this iteration.

Issue #484, opened 2017-12-20
Fix selection styles


michael:
Currently the dashed selection rectangle is display for regular selections as well. Restore previous style for regular selections while keeping the dashed lines for range selections while editing a cell.

Integral:
This issue was fixed by #500.

PR #497, opened 2017-12-21
Fix reference selection styles

Integral:
This is somehow fixing #484 and should be reviewed.
In my taste it's a bit hacky, but the alternative would be creation of selection overlay just for handling reference selection, i think it's a bit stupid since it's absolutely the same as a range selection, it's only need to be visually distinguished.
Maybe we should return to this question when we will need to expose multiple range/reference selections. Because this would be not enough to cover that case.

### DIFF
--- a/src/sheet/SheetComponent.js
+++ b/src/sheet/SheetComponent.js
@@ -216,10 +216,13 @@ export default class SheetComponent extends CustomSurface {
     }
   }
 
-  _positionRangeSelection(sel) {
+  _positionReferenceSelection(sel) {
     const rects = this._computeSelectionRects(sel)
     const styles = this._computeSelectionStyles(sel, rects)
-    this.refs.selRange.css(styles.range)
+    const selRange = this.refs.selRange
+    selRange
+      .addClass('sm-reference-selection')
+      .css(styles.range)
   }
 
   _computeSelectionRects(sel) {
@@ -377,6 +380,11 @@ export default class SheetComponent extends CustomSurface {
     }
 
     return styles
+  }
+
+  _hideReferenceSelection() {
+    this.refs.selRange.removeClass('sm-reference-selection')
+    this._hideSelection()
   }
 
   _hideSelection() {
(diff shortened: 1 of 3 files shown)


oliver----:
I fear that I am not fine with this implementation. Probably I have overlooked this in the previous PR already.
The original idea was to pass an extra overlay maintained by SheetEditor to the  SheetComponent. And let SheetEditor position that rectangle when it needs, using a helper method provided by SheetComponent.

Integral:
Are you suggesting to add new overlay and do the very the same thing what is doing the range selection overlay? Because the only difference now is dashed style.

oliver----:
It is an important detail, I think. Michael and me were discussing a lot how to improve the UI infra-structure, and one key thing is to try to maintain UI elements on a higher-level so that we have more control of how they are rendered and when.
Like here: https://github.com/stencila/stencila/blob/718e97a798521fab3c8ec7f603cf15887ebf704d/src/sheet/SheetEditor.js#L150

If you have this structure you can actually control it pretty easily. You can even solve the multi-rect case pretty easily.

Integral:
@oliver---- this is nice.
But we need to discuss it more.
It's the other, i believe not so small, task.

michael:
Yes I guess it would be good to let the parent component manage the selections and position them using an API provided by SheetComponent. However, when we have multiple selection rectangles adding/removing them would trigger a full rerender of the SheetComponent. Is that acceptable?

oliver----:
I think the re-rendering problem is a different topic (~e.g. selective shoudRerender, or incremental rendering)

oliver----:
Rather having this feature just 'nailed' I would prefer to have a good implementation :)

michael:
Idea: Create a dedicated component `SheetSelections` that is owned by `SheetEditor` and passed to `SheetComponent`. That way we have a more minimal re-render.


```js
// in SheetEditor
this.refs.sheetSelections.setProps({
  [ {type: 'range', rects: this.refs.sheetComponent.computeSelectionRects(sel1) ],
  [ {type: 'range', rects: this.refs.sheetComponent.computeSelectionRects(sel2) ]
})
```

oliver----:
Superseded by #500

PR #499, opened 2017-12-21
Cell overflows.

Integral:


src/sheet/_sheet.css
```diff
@@ -301,3 +301,11 @@
   text-align: center;
   font-size: 12px;
 }
+
+.sc-sheet-cell .sc-plotly-value {
+  position: absolute;
+  top: 0px;
+  left: 0px;
+  background: #fff;
+  z-index: 10;
```
oliver----: @michael is there some advice on how to use z-indexes?

PR #500, opened 2017-12-21
Cell references

oliver----:
Improved implementation according to discussion in #497


oliver----:
It's working like this now:
```
this.refs.cellRanges.setProps({
  ranges: [{ top, left, width, height },...]
})
```

While `SheetComponent` provides a helper to get a visible rectangle for something like
```
sheetComponent.getRectangleForRange({ anchorRow, anchorCol, focusRow, focusCol })
```

PR #504, opened 2018-01-22
Dynamically loading of functions

nokome:
At present only Javascript function libraries can be loaded and only at build (bundling) time. This PR allows for dynamic loading of libraries from external execution contexts (e.g. R, Python) in alternative languages.

The use case for this functionality is a researcher who would like to wrap existing R or Python code and make it available as a Stencila function. We are trying to keep this process as simple as possible and allow users to use existing documentation formats and tools. Allowing for "hot-loading" functions provides some motivation - allowing them to see their function immediately available - and for manual testing.

For example, here is an RStudio session which uses the [stencila/r](https://github.com/stencila/r) package to register some custom functions with Stencila:

- `register_function`: registers a function (either from source file or a function object) and converts the function's documentation (either from same source files, built-in, or by parsing function object's source) to Stencila `Function` XML.
- `register_library`: registers a folder of functions in `*.R` files

The user can then connect to the Stencila `Host` that was started in R, see the documentation, and use, those functions:

image

Then within Stencila the user can

image

This PR also removes argument checking when calling a function from `MiniContext`. Checking of call arguments was done based on a function's XML spec, specifically, the `<params>` element - a form of dynamic type checking before each call. There needed to be concordance between the spec and the implementation (the spec wasn't just extra documentation).

This introduced friction when writing functions because, often, the implementation would diverge from the spec. To make it easier for contributors to write functions we`re moving towards dynamic argument checking/dispatching and XML specs/docs based of language native tools (e.g. JsDoc for Js, roxygen for R, docstrings for Python).

In the future we will introduce more formal static type checking (based on the entire call graph, not just single call arg checking). But for now this change makes it easier to implement and use functions.

### DIFF
--- a/src/host/Host.js
+++ b/src/host/Host.js
@@ -49,6 +49,8 @@ export default class Host {
      */
     this._peers = {}
 
+    // TODO: consider instantiating the FunctionManager here, within this host, instead of in setupStencilaContext.
+    // ATM, its the only place that FunctionManager and used
     this._functionManager = options.functionManager
 
     if (!this._functionManager) {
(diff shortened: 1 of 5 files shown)


nokome:
@oliver---- : could you please review this; let me know if you want to discuss.

src/host/Host.js
```diff
@@ -49,6 +49,8 @@ export default class Host {
      */
     this._peers = {}
 
+    // TODO: consider instantiating the FunctionManager here, within this host, instead of in setupStencilaContext.
```
oliver----: It is also used for contextual function documentation

oliver----:
I agree.

PR #506, opened 2018-01-23
Update to latest Texture and work on examples

michael:


nokome:
To execute the R and SQL cells in the documents. Run the `stencila/core` image (this will download the latest version if you don't already have it):
```
docker run -it --rm -p 2100:2000 stencila/core
```
Then, as a temporary hack, instantiate an RContext in the running container (this is necessary because we want to force the use of the SQLiteContext that comes from the R package, and not the one that comes from the Python package which has old API - I will fix the container tomorrow)
```
curl -X POST http://localhost:2100/RContext
```
Then add the URL of the container to the end of the dev URL
```
http://localhost:4000/examples/index.html?example=publication&peers=http://127.0.0.1:2100
```

oliver----:
@michael looking good

nokome:
Merging this now so that we have a version on builds.stenci.la that works with the latest Docker image.

PR #512, opened 2018-01-25
WIP: Host interfaces

nokome:
Add user interface for Host. See StEP `0007-hosts-interface.md`


Integral:
@nokome, sorry for being so slow
here is something like first version
screencapture-localhost-4000-examples-index-html-1517225715762
there are questions indeed...
1) thing like http://localhost:4000/examples/index.html?example=external-cells&peers=http://127.0.0.1:2000,http://127.0.0.1:2100 will not open at all, if one of the peers is offline than everything got blocked by js error:
screen shot 2018-01-29 at 14 38 48
2) there is no internal host inside `this.context.host`
3) there is no information in `host.peer[].types` related to python or sqlite packages:
screen shot 2018-01-29 at 14 48 38
so I was unable to perform url matching for highlighting instantiated instance...
4) there is no `host.discover` method

nokome:
Thanks @Integral it's great to have this implemented! With regard to your questions:

1. I fixed connection errors with 2f1bc25
2. `this.context.host` *is* the internal host, I've added it to the list in 8a5e5c4
3. I've dealt with this in afbfbfbd7
4. I fixed peer discovery with 47f4be5

So it looks like this now:

image

Obviously the UI needs to be refined, but I think this a good start and would like to merge it now. The only question I have is currently the implementation for this is in `src/shared` but its seems like it would make more sense for it to be in `src/host`.

nokome:
@michael and/or @oliver---- could you please review this and merge it if considered OK.

nokome:
Thanks all. Merging this now. A further iteration on the UI will be required later.

PR #514, opened 2018-01-25
Transclusion

oliver----:
Fixes implementation to consider cells from different resources.

### DIFF
--- a/examples/data/publication/manifest.xml
+++ b/examples/data/publication/manifest.xml
@@ -1,7 +1,7 @@
 <container>
   <documents>
-    <document id="doc-1" name="Reproducible Document Stack" type="article" path="manuscript.xml" />
-    <document id="sheet-1" name="Ice Cream Sales" type="sheet" path="ice-cream-sales.xml" />
+    <document id="doc1" name="Reproducible Document Stack" type="article" path="manuscript.xml" />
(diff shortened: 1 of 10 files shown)


oliver----:
@michael I reverted to what I had yesterday and added some lines. sigh.

oliver----:
@michael can u try a bit? and merge if you are happy.

oliver----:
This also resolves a regression with colliding cell ids after we have introduced multiple docs.

oliver----:
stencila-transclusions

nokome:
@oliver---- nice! thank you!

BTW: improving that plot function is high on my to do list - I will get back to those related PRs over the next few days!

oliver----:
Hehe... I thought you would spot that ;)

michael:
This feels just awesome! :)

Some thoughts:

- would be nice to be able to follow such links, like command click takes you to the transcluded range in the sheet
- to transclude datasets it would be nice to somehow say, i want all rows for columns A:C, so when a new row gets entered in the sheet, this change is reflected in the document where the dataset is used
- as you already mentioned we need a smooth transition between the docs, and keeping the scroll position (will work on that)

michael:
👍 for merging

nokome:
In addition to @michael's suggestions, it would be great to be able to transclude the entire sheet ie dataset.

PR #529, opened 2018-02-07
Added jupyter notebook example

apawlik:
A simple Jupyter Notebook example of using Python 3 for data analysis (plotting using pandas and matplotlib). Based on Software Carpentry lessons.

Example is to demonstrate Stencila conversion abilities.


nokome:
@apawlik : thanks for these! The best place for them at the moment is in https://github.com/stencila/examples repo (use a separate subfolder for each). Could you create a PR over there please (if you don't have time, let me know and I can do it). I'll close this one.

PR #544, opened 2018-02-07
Dar storage

michael:
This PR enables storage of Stencila projects (Dar archives)! 🎉 

Notable changes:

- Update to Texture Alpha 4
- Read from and write to Dar folders
- Removed all examples except for one (kitchen-sink); see #545 
- Computed values for cells and repo figs are now stored in the XML
- Changed bundle destination from `bundle` to `dist`
- Moved `examples/data` to `data` (everything now has the same layout as in Texture -> less load on our brains)

When starting the dev server by default the kitchen-sink example is served via the virtual file system. Change the url to enable saving like so:

```
http://localhost:4000/?archive=kitchen-sink&storage=fs
```

Pressing `CommandOrControl+S` saves the document (e.g. to `data/kitchen-sink`)

@nokome can you check this and give green light for merge? You should then do an iteration on the examples, and make sure we have complete folders (including sheet.xml files in the data subfolders)

@oliver---- I had to disable XML pretty print because it currently eats CDATA content.


oliver----:
@michael Working now.

PR #547, opened 2018-02-09
WIP: additional examples

nokome:
## Why?

Because we need more/better examples for testing and demoing.

## What?

Adds examples with source files in external formats:

- `mini` : an article and a sheet using mini
- `py-jupyter`: an Jupyter Notebook in Python
- `r-markdown`: an RMarkdown article
- `r-sheet`: a sheet using R
- `starwars`: data from the Star Wars API imported from CSV

The XML files, including the manifests are created by 'importing' the folder using `stencila-convert`. e.g.

```
stencila-convert import data/py-jupyter
```


michael:
That's a perfect approach I think. Only thing I'm not sure about is mixing source and and converted files in one folder. Maybe better to have them stored separately and make the stencila convert tool more low level. Like this:

```
src-data
  starwars/starwars-people.csv
```

Then apply conversion:

```
$ stencila-convert src-data/starwars data/starwars -i csv -o dar
```

which creates:

```
data
  starwars
    starwars-people.csv.xml
    manifest.xml
```

In our make.js we could scan for all entries in src-data and run the appropriate conversions automatically. But I think it's important to put low level converters in first. `stencila-convert import data/py-jupyter` seems to quite opinioated already.

I think the stencila core toolchain should not know about conversion features, hence putting a src link into document entries in the manifest may be one step too far already. Again, we can discuss such things later when it comes to ad-hoc conversion but I'd like to avoid adding an attribute to the spec unless we are really certain it should be added.

michael:
@apawlik you sure you want to cover references in the conversion already? Seems the format is pretty unstructured. Will be hard to turn them into JATS-element citations. See spec: https://github.com/substance/dar/blob/master/specs/JATS4M.md#ref

apawlik:
@michael It's more a placeholder at the moment. @nokome and I have been looking at it but as you say it's not quite there yet. @nokome - should I remove this?

nokome:
I have updated the Jupyter notebook example and added support for Pandoc style citations/bibliography in the converters. This doesn't support the cite2c Jupyter extension but I don't think that should be too difficult (there is an issue for that https://github.com/stencila/convert/issues/14)

Issue #550, opened 2018-02-09
Lack of repro-fig-preview component


Integral:
If you insert figure reference and will try to click on it, you'll get `No Component registered with name repro-fig-preview`

michael:
That only happens in Stencila right?

Integral:
True

Integral:
Should be fixed via PR #553.

PR #553, opened 2018-02-12
Reproducible figure preview with label.

Integral:
Why: There is repro-fig component missed as described in #550  
What: Added component and styles, so now we have dropdown with labels


michael:
Good as a first iteration. I will sketch some better styles for the different xref edit dropdowns, but this is not high priority.

Issue #545, opened 2018-02-08
Managing examples


michael:
I'm about to enable saving for document archives and the examples are in the way as they have dynamically generated sheets, which divert from a regular load workflow.

Let's now move towards persisted and checked in xml files for sheets. We just need to be careful to have a good way to manage those as we will have to update all of them when we make schema changes etc.

I will start with just one default example (kitchen_sink) so we can start the read/write workflow and stabilise it. Then we bring back the other examples.

We should also work on a migration script which we can manually run whenever we update the schema to have all examples updated.

nokome:
I have already reinstated the sheets examples, and made some changes to the index and other examples in this PR https://github.com/stencila/stencila/pull/527. So could you work off that (perhaps by merging first) so those changes don't get lost.

I agree with having checked in xml files for sheets. My only concern is that in the short term (like next few days), we need working examples of sheets _with_ the bug fixes that are currently being made.

The best approach may be to decouple demo examples. I can do a little bit more work on the converters. Write demo examples in stable, external formats such as Markdown and Excel, and then convert them to Dar. Then changes to Dar schema only require changes in one place, in the converters.  I'll work on this is today and get back to you. 

This would allow you to go ahead with the changes you propose here, and for me to independently work on demo examples.

nokome:
Some suggestions after playing with the Dar storage branch:

- delete `/index.html` and `/examples/examples.html` and instead put that info at the top of the kitchen sink example  i.e. change it's title to 'Stencila Test Examples' and near the top have links to the other examples - that way the first page at http://localhost:4000 provides an entry point to the other examples but also has all the things in it that you would commonly want to test - and there are no HTML documents to maintain, just Stencila projects!

- personally, I'd prefer `/data` to be `/examples/data` just to reduce the number of top level dirs and because 'data' implies that the app embeds data

nokome:
Oh, just saw this in #544 

> Moved examples/data to data (everything now has the same layout as in Texture -> less load on our brains)

Fair enough, ignore the second bullet point above :)

michael:
Actually folder examples should be renamed to app.

michael:
Suggestion:

- Rename `examples` to `app`
- Rename `data` to `examples`

Would do that for both Texture and Stencila. What do you think?

michael:
[...]

PR #558, opened 2018-02-13
Reorganise examples

nokome:
## Why?

Better naming of directories. See the discussion in #545.

## What?

- Rename `examples` to `app`
- Rename `data` to `examples`
- Related changes in `make.js`
- Reduce number of `*html` files (still have a separate hand written `index.html` although that may later be rewritten as a Stencila document).


michael:
Ok, merging this, I can see how we want to link to multiple examples at the moment.

PR #560, opened 2018-02-13
Hosts panel.

Integral:
Why: There are broken styles now in hosts panel.
What: Fix styling issues.


michael:
Better. Could still need some alignment (like checkbox and input field). But let's focus on more important things first.

PR #566, opened 2018-02-13
Insert reproducible figure command.

Integral:
Why: There is no ability to insert reproducible figure. 
What: This PR adds command for inserting reproducible figure


oliver----:
Looking good.

PR #567, opened 2018-02-13
Connect global key handlers on app level

michael:
### Why

Some keyboard shortcuts were not working when no editor has a focus (e.g. CMD+F for bringing up a find and replace dialog).

### What

Solved this by connecting the global key handlers on Project / app level.


oliver----:
Yep.

PR #570, opened 2018-02-14
StEP: Engine (Iteration 1)

oliver----:
## Why?

Engine needs redesign.
see https://github.com/stencila/stencila/blob/step-engine/docs/steps/0008-engine.md

## What?

- better architecture embracing problematic scenarios
- fully test covered


oliver----:
We are getting closer :)

oliver----:
FYIW, I am replacing the Engine, by a fully-testable Engine. In this first iteration, I am trying to achieve bring back the current functioning. We will then have a better base to start fixing things (because it is now test-driven) and we will have a better base to enable things like validation (particularly type) and other stuff.

oliver----:
@nokome it would be great if you could start reviewing.
I'd recommend rather to do a general review on the source files, than on the commits.
I'd like to have this PR merged soon, so that we can start working with it the next days.
Thanks!

oliver----:
I will try to clean-up StEP and source documentation (Engine, CellGraph)
The two tests might be a good starting-point for you.

oliver----:
image

nokome:
@oliver---- : thanks! Will try to look at this later today (but have to finish an iteration on my own PR first).

PR #578, opened 2018-02-16
Documentation: FAQ and stab on Quick Start

apawlik:
Because it's branched from current `master` it does not have the Installation docs and side bar update (see https://github.com/stencila/stencila/pull/577) but these two should merge into `master` without conflicts.

Added FAQ and first stab on Quick Start (needs work).
Updated a few links 
Added a set of screenshots to be used for the documentation files


apawlik:
PRing to wrong branch. Will close and PR again

PR #582, opened 2018-02-19
Examples rmd

apawlik:
A simple example of Rmd with R and SQL code bits using sample data taken from Data Carpentry workshops


nokome:
@apawlik : I'm going to close this PR as I think that this would be better as an example in the https://github.com/stencila/examples repo. I won't delete the branch for now so we can come and grab it to put there.

apawlik:
Good idea. Thank you!

apawlik:
Done https://github.com/stencila/examples/commit/5620da5b06d250fc00d6953bd1155eccc16a2af1
I will delete the branch now

Issue #588, opened 2018-02-20
Selection rectangle gets messed up when editing a second sheet


michael:
Could again be some life cycle related problem.

Reproduce:
- Open kitchen sink example
- Open Ice Cream Sales sheet (and edit some cells)
- Create a new sheet (and edit some cells)

michael:
Can confirm this happens only when switching from one sheet to another.

michael:
Fixed.

PR #590, opened 2018-02-20
Fix sheet lifecycle

michael:


michael:
@oliver---- Life cycle is a bit tricky to handle with the double rendering in place. Maybe you want to have a look if you see room for improvements?

This fixes #588

PR #605, opened 2018-02-27
Remove display modes.

Integral:
We decided to remove display modes feature for some time and implement it later.
This PR removes everything related to concept of modes leaving only original "minimal" mode.

### DIFF
--- a/src/shared/_cell-value.css
+++ b/src/shared/_cell-value.css
@@ -57,7 +57,7 @@ plotly defaults
   text-align: center;
   display: table;
   table-layout: fixed;
-  min-width: 100px;
+  min-width: 80px;
 }
 
 .sc-table-value table thead {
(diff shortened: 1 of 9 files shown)


Integral:
@michael I also removed overflowing cells, however now we will have the same problems as before with plotly graphs. So let's not merge, but decide something and implement it.
Let me describe this problems. Currently we are setting default size of graph or expecting default size from libcore. So without overflow component it will take a height of graph, but the width would be still overflow... (we didn't set size before which results in ugly thin graphs).
So I think we should save that default size, but instead of setting it explicitly on graph we should rather set width of column allowing column and graph resizing (we should fix plotly rerendering issues as well then). What do you think?

michael:
I fixed this by setting overflow hidden of cells. Merging this now. We know this needs work, but for now this is the best outcome in terms of UX until we tackle sheet rendering.

PR #608, opened 2018-02-28
Cross browser type over.

Integral:
Current type over relies on experimental technology which is working only in chrome bigger than 60 and opera. 
This PR changes implementation to more classical one.


oliver----:
👍

michael:
Thank you guys!

PR #609, opened 2018-03-01
Remove typed cells from UI.

Integral:
We decided to remove typed cells from the sheet UI until we introduce issues features.
This PR removes type cell selection from the sheet toolbar and the column header settings widget as well as all related commands and helpers.


michael:
Can we solve this with just commenting out the commands, wouldn't want to kick out all the code, as we will bring it back.

Integral:
@michael, sorry, I thought we want to produce as much clean code as possible for now. But ok, I brought it back.

Issue #596, opened 2018-02-23
Errors are not cleared after delete of erroneous sheet cells


michael:

Integral:
All that I can said now that the reason behind this is that cell still have error state and sheet engine adapter is calling engine.removeCell.

Integral:
Proposed fix here: #610

oliver----:
This is particularly the case when changing a cell from an expression to a constant.
The engine should wipe errors when analysing constant cells.

PR #610, opened 2018-03-02
Clear errors after changing expression to constant.

Integral:
We have a bug (#596) with error notifications in sheets which are become immortal and you can't remove it neither with constant input, neither with deleting cell contents. The problem came from engine bug: when you are changing expression to constant cell got removed from graph, but the state didn't change for a cell. 
This PR addressing described bug by removing stuck error message, deriving cell status and notifying cell.


oliver----:
This seems to be fine now.

Issue #613, opened 2018-03-07
Improve Help Panel


michael:
For a start we could do something like this:

![image](https://user-images.githubusercontent.com/2931/37091199-93ab7588-2207-11e8-8e70-d930cd94dbea.png)

PR #614, opened 2018-03-08
Improve help panel.

Integral:
Style changes as suggested in #613.


michael:
👍

The separation line on the side panel is not pixel perfect but we need to tweak that anyways. Merging.

Issue #612, opened 2018-03-07
Simplified Runtime/Hosts Panel


michael:
![image](https://user-images.githubusercontent.com/2931/37090604-486c7c5e-2205-11e8-8447-546ecc786cd2.png)

- we use auto discovery for default hosts (e.g. Stencila hub + localhost)
- by selecting a host a user implicitly determines the runtime identifier (each host is exposing it) and we store it in the manifest.xml
- you can manually connect to a host using a separate dialog (there you enter a host url, including an API token etc.). this can be used to connect to custom images with custom libraries etc.
- clicking on the info sign opens a dialog with details about the the runtime (what languages in which versions, what libraries, which mini function libraries)

michael:
Ok, some more thoughts on this. It is actually very important to select a runtime explicitly, and store that identifier in the manifest. Those runtimes need to be registered and verified, so it is actually problematic if allow creating new runtimes from the user interface (as the project would then no longer be reproducible).

What we probably want is a select box to choose an officially supported runtime in Stencila (Stencila Core / Mega). Then we list available hosts that provide exactly that runtime (we need a way to ping a host and ask for it's runtime and version)

Visually it could look like this:

![image](https://user-images.githubusercontent.com/2931/37106516-5095e154-2232-11e8-916d-dd0457f907fc.png)

nokome:
This looks good and a big improvement of what we currently have.

I prefer "Environment" to "Runtime" - I think it is more accurate. I think of the container images (Docker, Singularity, whatever) encapsulating a computational environment and the "runtime" being the language runtime which is providing the HTTP server for the host (e.g. R, Node). That is the terminology we used over in stencila/images anyway.

One thing for us to consider is that users will be able to install Stencila language packages locally (e.g. https://stencila.github.io/r/) and run a host that is executing within their local environment (e.g. the R packages thay have installed). This will give flexibility for users early in project lifecycles and not require Docker or connecting to an remote host (closer to publication they could use an existing or custom image). Hosts need to include in their manifest the environment's that they support, including a possibly ill-defined local environment which might be in the list above as `local/local`. I'll work on this tomorrow.

michael:
With these local/local environments, will you be able to also use other languages, e.g. SQL? Like I install the R language package, will I get just an R environment or Stencila Core + my custom R?

We may run into the situation that we need to disable some mini functions, which are backed by other contexts (e.g. Python), right?

[...]

PR #619, opened 2018-03-26
Changes to remote Host API

nokome:
Changes to API for selecting and connecting to remote Hosts. This is required before progressing https://github.com/stencila/stencila/issues/612


nokome:
An initial iteration done on this including move towards UX/UI in #612.

peek 2018-03-27 16-52

nokome:
More refinements can be made to UI/UX but I think this should be merged now. Auto-discover is off and instead, in dev mode, you connect to a host using the `hosts` url parameter (in the desktop a connection to the embedded Node.js Host will be done at startup).

You can test this against the `master` branch of `stencila/node`:

```
git clone git@github.com:stencila/node.git
cd node
npm start
....
Host HTTP server has started:
  URL: http://127.0.0.1:2000
  Key: cf85bfe9e71bad413457c6bc75c6c48652d26856d8893875be34b68a44ff961a
```

and then add `&hosts=http://localhost:2000|cf85bfe9...` to the dev server UL. Or, turn auth off and just use `&hosts=http://localhost:2000`:

```
STENCILA_AUTH=false npm start
```

The `stencila/node` Host now provides a `SqliteContext` with an implementation of the 'data pointers' (as indicated by the ellipsis at the end of a table):

peek 2018-03-31 09-07

michael:
Just tried this out. Working nicely!

@nokome could you work on incorporating this into stencila-desktop? I see that we use a native extension for sqlite3. Don't know if this works smoothly with electron-builder app bundling. Please check. I'll update stencila desktop to use the latest stencila version and you should be good to plugin the node host there.

PR #623, opened 2018-03-27
Cell value refactor

Integral:
This PR contains cell/cell value components refactor as well as delay processing of errors.


nokome:
Nice👍

PR #630, opened 2018-03-29
Manual vs automatic evaluation of cells

oliver----:
# Why?

The original idea was to introduce breakpoints to give the user more control of cell execution e.g. in presence of long-running tasks.

# What?

For sake of simplicity, we introduce a flag on document level, that controls whether cells are evaluated automatically or triggered by the user explicitly.
A manual request for evaluating a cell is interpreted as 'run this cell and all its dependencies' (i.e. anything needed to run this cell).

### DIFF
--- a/src/article/ArticleAdapter.js
+++ b/src/article/ArticleAdapter.js
@@ -143,12 +143,12 @@ function _getCellData(cell) {
 }
 
 function _addAutorunFeature(doc, adapter) {
-  Object.assign(doc, {
-    set autorun(val) {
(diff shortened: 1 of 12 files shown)


oliver----:
@michael @Integral please try out.
We could get this merged, and continue with a general iteration of fixes.

oliver----:
but maybe let's get the tests green before we merge.

oliver----:
It seems that with `autorun` off cells are not analysed correctly.

PR #631, opened 2018-03-29
Use shared AppChrome implementation

michael:


oliver----:
Basically good. Because less redundancy.
Just realized that with the ability to add new documents ('+' button) we need a way to register the new resource in the same way as the initial ones...

oliver----:
Is this ready now?

PR #638, opened 2018-04-16
Function libraries and call semantics

nokome:
## Why

This PR is needed because of updates to the definition of function libraries in JS and other languages including:

- changes to how functions are called (allowing for variadic and extensible keyword arguments)

- allowing functions to be defined dynamically in external execution contexts (e.g. Node.js Python, R) e.g. in `libdh`

- changes in function definitions away from XML towards native syntax for each language (e.g. JSDoc)

For more details see  https://github.com/stencila/libcore/pull/20 and https://github.com/stencila/node/pull/30

https://github.com/stencila/libcore/pull/20 needs to be completed, merged and published before merging this.

## What

- mostly refactoring/renaming

- does add an undocumented 'hack' to allow functions to be defined within a cell - this is really useful for less experienced developers to test that a function and its documentation works. To be decided if this remains undocumented, is documented, or removed (and a perhaps function definition UI added instead). An example or defining a function definition in Javascript using the Node.js context:

image

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: node_js
 
 node_js:
-  - 6
+  - 8
 
 services:
   - docker
--- a/app/example.html
+++ b/app/example.html
@@ -9,8 +9,8 @@
     <script type="text/javascript" src="./lib/substance.js"></script>
     <script type="text/javascript" src="./lib/texture.js"></script>
     <script type="text/javascript" src="./lib/stencila-mini.js"></script>
-    <script type="text/javascript" src="./lib/stencila-libcore.xml.js"></script>
-    <script type="text/javascript" src="./lib/stencila-libcore.all.js"></script>
+    <script type="text/javascript" src="./lib/stencila-envcore.min.js"></script>
+    <script type="text/javascript" src="./lib/stencila-libcore.min.js"></script>
     <script type="text/javascript" src="./stencila.js"></script>
     <script type="text/javascript" src="./vfs.js"></script>
     <script type="text/javascript" src="./env.js"></script>
--- a/make.js
+++ b/make.js
@@ -3,7 +3,6 @@ const b = require('substance-bundler')
 const fork = require('substance-bundler/extensions/fork')
 const install = require('substance-bundler/extensions/install')
 const path = require('path')
-const isInstalled = require('substance-bundler/util/isInstalled')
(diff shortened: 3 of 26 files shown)


oliver----:
Nice!

oliver----:
@nokome better to rebase onto latest develop.
Your changes are orthogonal though, just better to have it moved to the latest state.

nokome:
OK, I think this is finally ready to merge. More work needs to be done on several things including

- function usage tooltip and help panel (showing alternative parameter type signatures, better rendering of help)
- `JsContext` call semantics (matching named arguments to parameter names)
- removing code that is now unused (e.g `FunctionDocument`)

But this iteration is compatible with the latest version of libcore which has better documentation and things like a variadic sum function (finally :tada:) - so we should merge it in for the next release :)

image

nokome:
Just been testing the integration with Stencila Desktop. Simulanteous, reactive calculation of a sum in different languages :roll_eyes: 

peek 2018-05-02 20-59

oliver----:
On the long run I would like to see this also available for the `js` context. Maybe we can address this together with a webworker'isation of the JsContext.

oliver----:
Couldn't get this running. Will let you finish the `stencila-desktop` integration first and then try again.

nokome:
I've merged this.

The main change from the screenshots above has been to treat functions as first-class objects that can be returned from cells. When a function is an output from a cell it is registered in a `local` library.

>  I would like to see this also available for the js context (in browser). Maybe we can address this together with a webworker'isation of the JsContext.

Sounds good. I was hoping we could have just one implementation of a `JavascriptContext` https://github.com/stencila/node/blob/master/lib/contexts/JavascriptContext.js#L10 which would be used here and as a base class for `Node Context`

One aspect of this PR to note is that the `Engine` is now a property of the `host`. Prior to this PR we were actually instantiating two engines (but only running one). I have kept the one inside the host because the host needs to create execution contexts and have them register libraries (including `JsContext`) before it starts the engine. 

I found the whole app setup/initialisation pretty confusing with multiple nested functions e.g. `setupStencilaContext`, `_setupStencilaChildContext` and it seems that it would be good to refactor and simplify sometime (at which pint we can decide onthe best place for the engine to live).

oliver----:
I agree, the setup is confusion at the moment. Let's improve this later on.

PR #651, opened 2018-04-30
Resize sheet on paste

oliver----:
# Why?

At the moment, when pasting a table into a sheet that is larger than the sheet itself, the data gets truncated. See https://github.com/stencila/desktop/issues/35

# What?

This PR introduces logic that automatically adds columns and rows as necessary to fit the pasted data.


nokome:
Am I right in assuming that removing WIP from the title means this is ready to merge @oliver---- ?

oliver----:
Right. This what I usually do.

nokome:
Thanks!

PR #654, opened 2018-04-30
Some styling and branding work

michael:
Using the Stencila blue, but not too much. E.g. we are using white
overlays now, as making these blue would look extremely acid-y.

Not sure if I'm entirely happy with the round buttons. We had them in Substance before, but they take more space and are hard to correctly align them (so they all look balanced/centered).

I've put those overrides into a separate `_branding.css` file. We'll be good with this for the release, but I'll reorganise this once we have settled on a design direction.

@nokome please take a look and merge when happy.


Integral:
I've added texture editor style override to prevent things like this:
screen shot 2018-05-02 at 10 59 02

nokome:
Thanks for doing this @michael and @Integral 

I think this is fine for now. But in the long term were are going to have to put some effort in to finding a design direction that everyone is comfortable with (if that is ever possible :))

The only thing I'd like to change, is to go back to the more minimal 'green dots' (or orange, or red depending upon state) that we used to have for the hidden code cells: e.g. https://www.youtube.com/watch?v=PCRoc0JxRWg

michael:
Yes, I'm aware the the styles are not perfect yet, which is true for Texture and Stencila. I'd try to share as much as possible to keep the maintenance effort as low as possible. Let's try to evaluate the best options and make decisions based on arguments (like user tests, good contrast, readability etc.).

I'll go back to the more minimal green dot for hidden code cells.

PR #657, opened 2018-05-03
Valid resource names

oliver----:
# Why?

At the moment, we do not prevent resources having the same name. This name collision is however a problem when it comes to transclusions.
Stencila should 
1. prevent the user from creating duplicate names
2. not create duplicate names as well
3. transform broken archives during import, i.e. archives with duplicate resource names (result should lead to visible errors that need to be fixed by the user)

# What?

This PR 
- introduces validation for the input used by the `ProjectTab` for editing the resource name.
- generates valid i.e. not duplicate names when adding Sheets/Documents
- appends '(duplicate)' to conflicting resource names during import

- Fixes https://github.com/stencila/desktop/issues/42
- Fixes https://github.com/stencila/desktop/issues/43


oliver----:
@Integral you might want to take a look at the Project test I have created here. This is the way how do minimalistic UI testing.

Integral:
I will! Thank you!

oliver----:
@nokome this is ready for review

Issue #659, opened 2018-05-03
Introduce Preview/Publish workflow/view


michael:
When a publication is ready for publishing, we want to see how it looks like in the final published version. Also the author may want to tweak several things:

- which code cells should be expanded (by default they are closed and can be revealed by the reader)
- select an online service (runtime) that should be used when the publication is reproduced by the reader on-site.
- include or not include a download link for the dar archive in the publication
- ...

The idea is that the author can go into a "publish" mode to see the page like as the reader would do. Now they can toggle or untoggle code cells and we will remember those settings in the dar (so you don't have to redo it every time you want to publish). Also the user will be able to set some other settings (and see it previewed instantly). The goal here is that we don't mix the tasks of authoring and publishing a reproducible document. In edit mode, we provide access to everything that's there. It does not make sense to hide code while authoring, as you may not understand the program then. Also things like ctrl+f to search for some code you entered will cause unintended results.

I'll come up with some mockups for this, but will do that after the release. We also need the static site-generator in place first.

PR #660, opened 2018-05-03
WIP: Remove ability to show/hide code in the editor.

michael:
After our call yesterday we discussed how we want to handle toggling of code cells (green dot). We've came to the conclusion that we want to handle these as presentation settings as part of generating the publication. That means we remove all show/hide code functionality from the editor and move this into a separate "preview" or "publish" workflow.

1. This pull requests removes the show/hide code functionality
2. Work on preview/publish dialog (see #659)

@nokome are you ok with this approach? It means we drop the green dot for now completely, and bring it back for the reader view on a follow-up release.

### DIFF
--- a/src/article/ArticleEditorPackage.js
+++ b/src/article/ArticleEditorPackage.js
@@ -20,9 +20,9 @@ import ImageValueComponent from '../shared/ImageValueComponent'
 import PlotlyValueComponent from '../shared/PlotlyValueComponent'
 
 import {
-  SetLanguageCommand, ToggleAllCodeCommand,
-  HideCellCodeCommand, InsertCellCommand, InsertReproFigCommand,
+  SetLanguageCommand, InsertCellCommand, InsertReproFigCommand,
   ForceCellOutputCommand, RunCellCommand
+  // ToggleAllCodeCommand, HideCellCodeCommand,
 } from './ArticleEditorCommands'
 
 import FunctionUsageCommand from '../shared/FunctionUsageCommand'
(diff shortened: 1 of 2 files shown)


michael:
Some show/hide code functionality may be useful in the editor too (cmp. code folding in code editors). However I find your proposed solution too implicit, if all the code is hidden, you may find yourself in a dead-end situation as a user. Same is true for hide abstract but there it's not so critical. My suggestion would be to leave out code folding completely in this first version, and wait for user feedback.

Imo the proper solution would be code folding, as done in Atom etc. On find+replace (you'd unveil the hidden sections one you navigate through the matches). This takes a bit of effort to implement though, not sure it's worth at this early stage, as there are other priorities.

nokome:
I'm still in favour of keeping code hiding at this stage (at least until some preview mode is available). I think the benefits of code hiding (mainly for demonstrating that these are just like normal articles but with code embedded in them) outweigh any potential UX problems.

But happy to be outvoted on it: @oliver---- , @Integral, @apawlik : what do you think?

michael:
> mainly for demonstrating that these are just like normal articles but with code embedded in them

the problem I see is that we are mixing up the discussions. The conversation about the editor should be independent from demonstrating publishing features I think.

So when it comes to features of the editor, i'd like to ask the question: Does this feature improve editing of repro-docs?

I'm hoping that I have a first version of the static site generation ready in a few weeks, so then we could demo the real-world workflow in an update. Also possibly publishing via Stencila Hub.

apawlik:
I'm in favour of all code hiding / showing available in the editor for now. It does differentiate Stencila allowing for very user friendly and not scary WYSIWYG look but still retains the features sought by the coding users.

michael:
I understand, so during editing we have two use-cases:

- code focused editing
- prose-focused editing (to polish for publishing)

To support that, we can leave the hide/show code functionality in. What do you think about displaying the green dots, just to indicate there is something.. and if you click on one of them, we could show all the code cells instead of toggling individual ones?

nokome:
> displaying the green dots, just to indicate there is something.. 

Yes (or red, or yellow depending on state), like we have, just different visual representation.

> and if you click on one of them, we could show all the code cells instead of toggling individual ones?

With the work flow described above, I feel this would be frustrating. I'd have to go to the top level menu item to hide code again. I think i'd just prefer to toggle on/off each cell - like we currently have. While it's always nice to remove a menu item (from the cell menu), I think that would come with a UX cost.

michael:
We will not remove the functionality. Changed the display to minimal dots (green, yellow, red).

oliver----:
@michael can we delete this branch?

PR #661, opened 2018-05-03
Display language used in a cell.

michael:
@nokome please review.


nokome:
@michael : looks good, thanks!

Issue #641, opened 2018-04-18
Improve UX of function help tooltip


oliver----:
At the moment the tooltip is shown whenever the cursor is on a function or inside the parameters.
This is basically good, however, it is a bit obtrusive with that popup being visible all the time.

Alternatives:
1. Only show the tooltip when cursor is on the function name.
2. Only show the tooltip when idle
3. Let the user decide to turn off automatic popups and trigger explicitly (e.g. as it is in Elipse using `Ctrl+Space`)
4. Only show tooltip when typing.
5. Disable the tooltip on arguments for functions that span over multiple lines.

oliver----:
Added alternative `4.`

apawlik:
Hmm, I kind of like 1, actually :-) I think it is similar to some other applications (cannot recall any at the moment...but it seems familiar). 
3. would be good but I am not sure what is the awareness among our target users that there actually **is** an option to turn the popups off.

oliver----:
Right, with `3.` the problem is really that in anyway it is a bit obscure to new users.
In Sublime it is a combi of `4.` and `3.`, i.e. the popup shows automatically while you type, otherwise you can trigger it using `Ctrl-Space`.

michael:
I most like `3.`, as it gives full control and zero noise. The only problem is that you'd need to learn a shortcut, which means many users may miss the functionality. I'd definitely go for 3 and maybe combine it with `1.`.

Integral:
Yeah, those things might be kind of a revelation. For instance I didn't
know about this shortcut in Sublime. So we need to find a proper way to
tell users about this thing. Maybe we should print combo while user is in
1. state?

nokome:
My preference is to implement `1` and then maybe add `3` later.

michael:
@Integral can you implement `1` and make a PR?

Integral:
Another option, we could switch off tooltip for a big arguments or multiline arguments. @oliver---- what do you think?

oliver----:
Personal I am in favor of a more consistent solution. But I will add your suggestion to the alternatives :)

PR #662, opened 2018-05-03
Only show the tooltip when cursor is on the function name.

Integral:
Why: We want to make a function helper tooltip less obtrusive.
What: Number of alternative strategies when to show a tooltip were suggested here #641.
The first take is to show it only when a cursor is on the function name.

### DIFF
--- a/src/shared/FunctionUsageCommand.js
+++ b/src/shared/FunctionUsageCommand.js
@@ -36,7 +36,11 @@ export default class FunctionUsageCommand extends Command {
 
     let candidate
     nodes.forEach((node) => {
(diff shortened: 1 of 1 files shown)


nokome:
Unfortunately, it is now not showing when you first write a function i.e. it should pop up at the opening parenthesis e.g. `add(`

peek 2018-05-04 09-45

Integral:
I see what you mean, thanks, will fix that

oliver----:
Nice and simple solution 👍

oliver----:
Ok.

oliver----:
@Integral pls rebase onto latest develop.

michael:
Taking over, will do the rebase.

michael:
It's a pretty good solution I think! Thank you @Integral.

Merging.

PR #664, opened 2018-05-04
Function values

nokome:
# Why

Some minor fixes to deal with display and registration of functions returned from cells.

# What

Allow for display of extended types (e.g. functions); unpack functions before registering

peek 2018-05-04 19-51

### DIFF
--- a/src/contexts/ContextHttpClient.js
+++ b/src/contexts/ContextHttpClient.js
@@ -69,9 +69,9 @@ export default class ContextHttpClient extends Context {
       let value = post.outputs && post.outputs[0] && post.outputs[0].value
       if (value) {
         if (value.type === 'library') {
-          this._host._functionManager.importLibrary(this, value)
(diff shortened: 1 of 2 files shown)


oliver----:
👍

nokome:
Thanks, I'll merge this then now. Yes, it needs the latest version of stencila/node which will be in Stencila Desktop.

michael:
Ah so this means you can't write client-side JS functions only node ones?

oliver----:
Not yet: see https://github.com/stencila/stencila/pull/638

nokome:
Hah, @oliver---- you beat me to it! Yes, specifically https://github.com/stencila/stencila/pull/638#issuecomment-386527149

Right now the `stencila/node/JavascriptContext` handles functions but nothing else, and the `stencila/stencila/JsContext` does everything else. We need to merge the functionality into one. I was planning to do that today but got caught up with a bunch of other fixes :man_shrugging:

PR #668, opened 2018-05-04
Better message for some errors and Engine fixes

oliver----:
# Why?

Currently some cell errors are not very revealing. E.g.
- 'Unresolved inputs.' when one or more inputs could not be resolved
- 'Cyclic Dependency' when there is a cyclic dependency

# What?

Being there already in the error instances, this PR introduces a cell error message helper that exposes this information.

- Additionally I have fixed some bugs in the Engine related to cell updates.
- And added tests for `cellHelpers`.

On the long run we would like to have even more, such as being able to jump to the source of errors.

### DIFF
--- a/src/article/CellComponent.js
+++ b/src/article/CellComponent.js
@@ -1,7 +1,7 @@
 import { NodeComponent, FontAwesomeIcon } from 'substance'
 import ValueComponent from '../shared/ValueComponent'
 import CodeEditor from '../shared/CodeEditor'
-import { getCellState, getError } from '../shared/cellHelpers'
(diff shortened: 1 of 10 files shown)


oliver----:
stencila-better-message-unresolved-inputs

oliver----:
stencila-better-message-unresolved-inputs

oliver----:
stencila-better-message-cyclic-deps

oliver----:
@nokome pls review and merge if content

oliver----:
I will add some tests to have a greener test coverage result ;)

oliver----:
I gonna merge :)

Issue #665, opened 2018-05-04
Error message goes away if cell code editor is blurred


michael:
I think we should always show the error message. Thoughts?

oliver----:
How does that look like?

michael:
![no-error-on-blur](https://user-images.githubusercontent.com/2931/39620592-06034136-4f8c-11e8-9f2c-c1f2031af828.gif)

oliver----:
Right, that is strange. @Integral do you have an idea how to fix this.
I think it happens, because the Cell is rerendered because of loosing the focus.
And the current renderer hides the error too eagerly (originally intended to reduce obtrusive error message while typing)

Integral:
Need to investigate, you can leave it to me and I will look at this in the
evening.
Re too eagerly, would be ok if I will increase timeout? So it will stay a
bit longer?

oliver----:
Re timing: Alright.
Re eagerly: No. It should detect if the cell error/status has changed, which is only happening if the user actively updates the source. Just re-rendering with everything being the same should not change the visual appearance.

oliver----:
The originally intended idea was:
"When the user is typing, show errors with a delay, but valid results instantly"

oliver----:
So we need to come up with an interpretation of "when the user is typing". My suggestion is to simplify this to "when the cell state is changing".

Integral:
E.g. so every time when the cell state is changing we will reset timeout
and start counting until 1 second for instance. So if nothing will change
we will show error (if it's there). Right?

oliver----:
Sounds good :)

Integral:
@oliver---- I have a feeling that it's exactly what we are doing currently :) I will look into that

oliver----:
I don't think that the cell state changes when changing the focus.

Integral:
It's a bug definitely!

oliver----:
This will be fixed via #672

PR #669, opened 2018-05-04
Cell errors

Integral:
Why: as were reported in #665 errors messages are gone after focusing other node. There were also considerations that rerender hides the error too eagerly.
What: this PR blocks cell rerendering in case of focus changing. It is also increasing error delay. So now it will show error only if you will not change the cell contents during one second (it was half a second before).

### DIFF
--- a/src/article/CellComponent.js
+++ b/src/article/CellComponent.js
@@ -218,7 +218,7 @@ class CellComponent extends NodeComponent {
           if(errEl) {
             errEl.setStyle('visibility', 'visible')
           }
-        }, 1000)
+        }, 500)
       } else if (status === OK) {
         this.oldValue = cellState.value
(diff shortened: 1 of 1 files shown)


oliver----:
@Integral this is definitely more 'direct' than I thought. Do you think this is sufficient?
I.e. maybe there are more cases, where the cell state does not change while an update comes in?

oliver----:
I don't think the increase delay is making it better.
@michael @nokome what is ur opinion?

oliver----:
@Integral your focus trick is working

oliver----:
Overall, I think we should do a dedicated iteration on the error UX later on and continue with the current implementation for now.

oliver----:
This is how it looks now:
stencila-better-message-cyclic-deps

PR #670, opened 2018-05-04
Upgrade bundler

oliver----:
# What

Upgrading substance-bundler with all tooling in latest state.


oliver----:
The updated package-lock gets rid of the security warning we have seen before.

PR #671, opened 2018-05-05
Simplify mini syntax error messages.

oliver----:
# Why?

Syntax errors in mini are overly complex.

# What?

This PR turns them into a simple 'Syntax Error', which is apparently less revealing, but also less obtrusive.


michael:
Good to merge?

oliver----:
@nokome are you fine with this simplification?

nokome:
Thanks for this. I think this is fine for now, but as discussed previously we need to do another iteration which focuses on error messages (and linked help) and display (e.g. line and column indication) across all languages.

PR #672, opened 2018-05-05
Improve error display

oliver----:
# Why?

#669 didn't work after all. It was introducing a regression related to rerendering of the ellipses menu.

# What?
This PR tries to simplify the implementation for delayed error display, so that is closer to what I intended initially.


oliver----:
stencila-better-error-display-1

oliver----:
stencila-better-error-display-2

oliver----:
Still it could be less jumpy to my taste... any ideas?

oliver----:
As a comparison: the same implementation with 'display:none' instead of 'visibility:hidden'
This means, that an error is expanding only after the delay, and collapsing during typing.

stencila-better-error-display-3

oliver----:
I like the latter better.

oliver----:
Not there yet. Found some glitches.

Integral:
I don't like this jumping effect at all. That's exactly what we did try to prevent during last session... And we came to that solution with visibility switching so you kind of switching between value and the error without jumping (at least for a single line value/error).

Integral:
@oliver---- it is still have a problems... just try first code cell in repro-pub fig example, try to append numbers:

oliver----:
Right. Also there is a flickering when blurring.

oliver----:
But getting closer.

oliver----:
IMO this is good now.

Integral:
@oliver---- I think it is totally ok now.
One thing I notice is that special symbols like ö throw errors in browser while cell stays green, however this most be the other issue.

oliver----:
@Integral Thanks!

Re exception with special symbols: can't reproduce this. Could you maybe open an issue with screenshots of the error?

oliver----:
Merging this now.

Issue #677, opened 2018-05-07
Fatal error when switching between documents


nokome:
This error occurs when you start writing a function in an article, and then switch to a sheet. For example you start writing `plot(Data!A...)` and then switch to the sheet to see what the range should be.

![image](https://user-images.githubusercontent.com/1152336/39684339-4874e1a2-520f-11e8-912c-4ad249241c85.png)

My experiments suggest that this only you have an **unfinished** function call in the article cell e.g. `sum` and `sum(1,2,3)` are fine but `sum(` will fail.

This is a 'fatal' error because it prevents the user from being able to switch back to the article - which in the desktop means you have to close the app and reopen the document.

oliver----:
@nokome this does not happen for me (on latest stencila@develop)
Any idea how to narrow this down?

oliver----:
Anyways, technically this happens when the selection points to a none existing surface.
I will add a guard to the respective code in substance.

oliver----:
@nokome If you can reproduce the error, could you try if #678 fixes the issue?

nokome:
Yes, fixes #678 fixes the issue? Thanks!

PR #678, opened 2018-05-07
Fix issue 677

oliver----:
# Why?

see #677

What?

While there might be another root of the problem these changes make the implementation more robust. 
See also https://github.com/substance/substance/pull/1184


nokome:
Thanks @oliver---- :+1: . Confirming that this fixes the bug.

PR #679, opened 2018-05-08
Upgrade Texture

oliver----:
# What?
- Integrating latest Texture
- added a save hook for the local browser version the checks if a saved archive could actually be reopened

### DIFF
--- a/app/app.js
+++ b/app/app.js
@@ -1,11 +1,28 @@
-import { getQueryStringParam, substanceGlobals, platform } from 'substance'
-import { StencilaWebApp } from 'stencila'
+import {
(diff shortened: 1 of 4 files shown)


nokome:
Just to let you know, I've tried this out in Stencila Desktop using `npm link stencila`. And seems to be working - some bugs and console errors, but the basics are working OK:

image

oliver----:
We will fix the issues with serialisation and then it should be ready.

oliver----:
We have started an integration test suite in Texture. From now on we should see regressions more easily: https://github.com/substance/texture/pull/520

oliver----:
@nokome I found another regression in Texture I need to fix.
Working on it.

PR #682, opened 2018-05-11
Integrate with JupyterContext

nokome:
## Why?

So we can try using `JupyterContexts` in https://github.com/minrk/jupyter-dar ! See the related PR 
https://github.com/stencila/node/pull/42

## What?

- [x] Add the option to set a cell's language to "Python Jupyter" (in Sheets and Articles)
- [x] Host can handle a request to create a `JupyterContext` with options `{language: 'python'}`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "stencila",
-  "version": "0.28.1-preview.2",
+  "version": "0.28.0",
   "description": "Stencila is the office suite for reproducible research.",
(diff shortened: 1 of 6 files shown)


nokome:
I'm going to merge this in now since it is a relatively minor change but allows users to test out the integration with Jupyter kernels. The list of languages in the cell menu is getting long and we might think about how to reduce it based on which execution context are currently available. But that's another iteration.

PR #697, opened 2018-05-16
WIP: A unified Javascript context

oliver----:
# Why?

At the moment, we have two different implementations `JsContext` and `JavascriptContext`. The former used in the browser, the latter as a base-class for the `NodeContext`.
`JavascriptContext` is the newer more feature rich implementation which we have extracted into its own repository now.

# What?

- Replacing `JsContext` with `JavascriptContext`


oliver----:
This is already working kind of with js#browser-integration.
Now I need to think about how we bring dynamically created functions and the Engine together.
At present, changing the function will not trigger an update of cells that are calling the function.

oliver----:
stencila-dynamic-functions-preview1

oliver----:
Note: the reason why I named it `JavascriptContextClient` here, is that I didn't want confusion with the original implementation and also because it will be soon a proxy to a web-worker.

oliver----:
Things are moving and are broken with the latest versions of `stencila-node`. See https://github.com/stencila/node/issues/43

I would very much like to hold this PR back until development in `stencila-node` et al. has stabilised.

PR #702, opened 2018-05-23
Avoid invalid calls into Prism

oliver----:
# Why?

We have seen some errors in the tokeniser used for syntax-highlighting. See https://github.com/stencila/desktop/issues/54 ( item 3. )

# What ?

This PR adds a guard  that checks that the tokenizer is actually available for a given language, before calling into Prism, and then fails gracefully by skipping the tokenization and displaying an error message on the console.
While the source of the error seems to be a wrong language identifier, which we need to investigate further, this still makes sense for sake of robustness.

### DIFF
--- a/src/shared/analyseCode.js
+++ b/src/shared/analyseCode.js
@@ -74,9 +74,13 @@ Prism.languages.insertBefore('sql', 'punctuation', {
 })
 languages['sql'] = Prism.languages.sql
 
-
-function tokenize(code, lang) {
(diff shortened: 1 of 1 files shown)


oliver----:
@nokome this should prevent us from seeing the 'rest' error.
With the error message on the console we should maybe be able to understand better what the origin of the error is.

nokome:
Thanks a lot for fixing this so quickly @oliver---- Will merge now

PR #728, opened 2018-11-16
Update the introduction-rds example

nokome:
## Why: Please describe why you are suggesting the PR.

The example needed to be updated with 'real' R code, using real data from the Lewis et al paper.

## What: Elaborate what your changes do, and how they address the problem described.

This involved:

- removing example data and Mini code
- inserting R code for Figures 1 and 2
- for Figure 3 which requires a larger image (otherwise the results get squashed) I took the approach of running the coude in an external files `figure3.R` and then making it a static figure with the `figure-3.png` that is generated.
- adding the CSV files which the example uses.

This example, with a fully configured Binder image (which installs necessary packages etc) is here: https://github.com/stencila/examples/tree/master/elife-30274. See the Binder button there for launching it on Binder.


nokome:
Merging this now so that it gets deployed to http://builds.stenci.la and we can test connecting it to Binder via Bindilla.

PR #730, opened 2019-05-31
feat: Added handler to log from logga to winston

beneboy:
Added stencila/logga to package.json then added a handler to send logga messages to winston.


nokome:
Thanks!

PR #778, opened 2020-06-17
ci(Docs): Add command for publishing Help docs to Intercom

alex-ketch:
- [x] Command for publishing help files to Intercom
- [x] Run command as a GitHub Action when commits land on master
- [x] Create a `Stencila Bot` account on Intercom to use as article author?
  - [x] Append list of authors found in article front matter to the end of the generated HTML document, linking usernames to GitHub profiles

---

## Things to note:

Encoda strips custom CSS classes from anchor elements, therefore I've added a rudimentary HTML manipulation heuristic which assumes that paragraphs which only contain a single link element should be treated as centred buttons.

Currently the author id used in the payload is set to @colettedoughty's Intercom ID. This is configured in the GitHub repository `Settings > Secrets`. Depending on account limits, we could set up a Stencila Bot account in Intercom and use that account's id.


alex-ketch:
I've addressed your comments @nokome. Merging and seeing if the GitHub action succeeds or not 🤞

PR #873, opened 2021-03-31
chore(clippy): Remove unnecessary return

timClicks:
Pretty minor, but I guess that's a good place to start..


nokome:
Thanks for this!  Minor, but nonetheless useful 🙂

PR #874, opened 2021-03-31
perf(cli): Compile JSON schema validator once

timClicks:
I wasn't sure which feature gate to put this behind, sorry. I noticed that the validator was being compiled at every call to `validate()`. This PR changes that behaviour to cache the validator once it's first been triggered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,22 @@
+# [0.45.0](https://github.com/stencila/stencila/compare/v0.44.0...v0.45.0) (2021-03-31)
+
+
+### Bug Fixes
+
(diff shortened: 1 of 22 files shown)


nokome:
Thanks for this @timClicks. I decided to use `once_cell` instead of `lazy_static` e.g.

https://github.com/stencila/stencila/blob/29974047175f41ac94167d79b54beecbe8d80ce4/rust/src/plugins.rs#L28

Would you be able to refactor to that so that we don't need to introduce another dependency?

timClicks:
Absolutely. Adding another dependency did feel very uncomfortable.

nokome:
Thanks 🙏🏽 !

PR #886, opened 2021-04-27
Start of a Desktop client

alex-ketch:
This PR adds the skeleton for an ElectronJS based desktop client which will be used to build up functionality in future PRs.

Outstanding items I'd like to address before merging this PR:
- [ ] Configure a CI job [in a separate file](https://stackoverflow.com/a/62416523/5686014) to build, test, release the app only if [code in the `./desktop` directory has changed](https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=schema%2Cparameter-schema#pr-trigger).


nokome:
Thanks for this Alex 👏🏽.

I think it is wise to only trigger tests and builds when needed but think that we need to apply that approach across the whole repo, not just the `desktop` folder. 

I also think that we need to consider the dependencies between folders. For example, if someone makes a change to `cli` or `help` then there is no need to run CI for any of the other folders because none of them depend on those. However, if I make an API change in `rust` it could potentially break `node` and therefore `desktop` - I think new should know about that without having to wait until there is a change in either of those dependent folders.

So I think we should use a combination of the Azure Pipelines `dependsOn` and `conditions` clauses and analysis of git commits like [this](https://pumpingco.de/blog/run-an-azure-pipelines-job-only-if-source-code-has-changed/) and [this](https://pumpingco.de/blog/run-an-azure-pipelines-job-only-if-source-code-has-changed/) to trigger what gets tested / built on each CI run. I suggest that we develop that in a separate PR to this.

alex-ketch:
Merging this PR as discussed with @nokome, with above mentioned CI work to be added in future PRS.

PR #891, opened 2021-04-30
Various improvements to `help`

nokome:
Various things needed before publishing to help.stenci.la.


nokome:
@alex-ketch I've spent quite some time trying to use MP4s for the CLI demos (the GIFs were getting very large and time consuming to build). I jumped through various hoops but did not get satisfactory results.

In any case, the primary benefit of using Asciinema to record the screen casts is that it produces high fidelity "console-casts" without heavy payloads (and with the benefit of being able to copy paste bits of text). So it seems to be a shame to introduce a bunch of complexity in the build just to loose these benefits.

Instead, I think we should try using the https://github.com/asciinema/asciinema-player which would allow us to playback the `.cast` (JSON files) directly. However, I'm unsure of how to add the script and css to the docusaurus site. There is https://docusaurus.io/docs/docusaurus.config.js/#scripts but we probably only want to use that script on the demo pages. Would you mind having a look into that?

PR #929, opened 2021-05-16
Generate TypeScript types from JSON Schemas

nokome:
This PR adds the generation of `node/src/types.ts` from the JSON Schema generated by the Rust `projects`, `plugins` and `config` modules. That has necessitated some changed to the JSON Schema which may affect the desktop config views:

- sub-schemas are now all in-lined which may obviate the need for the following line and it's associated dependency
- config schemas no longer have a `title` (this is to avoid them having awkward and long TypeScript type names) - instead the first line of the description should be used as a "human readable" title.

https://github.com/stencila/stencila/blob/d0305948cd630a95de96023d66a829627d7745d4/desktop/src/main/config/index.ts#L11

See the `.test.ts` diffs for more details in the change of the schemas.

Currently, all the config types are nested within `Config`. It is easy to split them out into their own types - see [this](79911e3b44aeca59c2a74902ef7e8ccd017f9060) commit for an example of that (for `File`)


alex-ketch (commented):
Thanks for this @nokome 🙏 
I've got a question and a requested change/addition:

**Q:**
Not sure I quite understand the reasoning behind omitting the `title` field from the Config schema. As it stands I'd rather not have the title appear in the `description` at all and use the object `key` as the title if needed.

**Feedback/change:**
Could you please export the `types` from the project index file please? Currently they are inaccessible without being aware of the project file structure.
Unless there are some internal/private types we don't want to expose, the following has been working for me as I develop the UI:

```diff
# /node/src/index.ts
export * as config from './config'
export * as logging from './logging'
export * as plugins from './plugins'
export * as projects from './projects'
export * as pubsub from './pubsub'
+ export * from './types'
```

nokome:
> Not sure I quite understand the reasoning behind omitting the title field from the Config schema.

The appropriately titled `json-schema-to-typescript` package uses the `title` as the name of the generated interface. Furthermore, any subschema that has a `title` was getting it's own interface, rather than being inlined. That meant that it was generating lots of enums and interfaces with names like `interface LoggingToStandardErrorStream`. The net result was, IMO, a messy set of types with poor developer ergonomics.

> As it stands I'd rather not have the title appear in the description at all and use the object key as the title if needed.

I'm happy to hack something into the build process so this achievable. It does make more logical sense to have a human readable `title` and a separate type `description`. "Shouldn't be too hard" (Bentley, 2021. _Famous last words of scientists turned software developers_).

> Could you please export the types from the project index file please?

Sure, done!

PR #934, opened 2021-05-18
Desktop: Projects UI

alex-ketch:
This PR adds the preliminary UI for working with Projects, and subscribing to project file update events.

@nokome, there seems to be an issue with the `files` object returned when a file is deleted.
The deleted file is still present in the object, while the file` key is `null`.
This means that new files added to the project automatically show up in the sidebar, but deleted files are not removed.

**UPDATE:** Seems like a bug in the file deletion event. Instead of a `deleted` event, it sends a `created` event:
```ts
{
  "file": null,
  "files": {
    …,
    "/Stencila/Demos/Pastoll et al/Functions copy.R": { … }
  },
  "kind": "created",
  "path": "/Stencila/Demos/Pastoll et al/Functions copy.R",
  "project": "/Stencila/Demos/Pastoll et al"
}
```

Screenshot 2021-05-18 at 16 09 04


nokome (approved):
Thanks Alex. I've made some fixes on the Rust side and will merge.

PR #970, opened 2021-06-03
Use document ids instead of file paths

### LINKED ISSUE #968: Rust: document's should be keyed by unique id rather than path
Currently `Documents::registry` is a `HashMap<PathBuf, DocumentHandle>` using the absolute path of the document. This is problematic if the file is renamed or deleted - amongst other things it causes a panic in `Documents::get` when attempting to `canonicalize()` a path that no longer exists.

Instead, I propose to generate a unique id for each document that is the SHA-256 of it's _relative_ path within the project when it is opened.  That would allow the document to be renamed or deleted without problems.

Furthermore, by using the SHA-256 of the relative path we can store pipeline metadata about a document (e.g. upstream files that it was generated from) within a `.stencila` folder. (There are potential issues if the document is renamed externally but there are ways of dealing with this e.g. [like Git handles renames](https://chelseatroy.com/2020/05/09/question-how-does-git-detect-renames/))

nokome:
This closes #968. There are a number of issues with using the `path` as a unique identifier for an in-memory document (e.g if the document's file is renamed or deleted, or it is "saved as").

@alex-ketch I've renamed a lot of args, variables, functions etc in the `desktop` from `filePath` and `path` to `documentId`. Most importantly getting and using the document's id when it is opened:

```ts
  ipcMain.handle(
    CHANNEL.GET_DOCUMENT_CONTENTS,
    async (ipcEvent, filePath: string) => {
      let documentId = documents.open(filePath).id
      documents.subscribe(documentId, ['modified'], (_topic, docEvent) => {
        ipcEvent.sender.send(CHANNEL.GET_DOCUMENT_CONTENTS, docEvent)
      })
      return documents.read(documentId)
    }
  )
```

That works mostly. However, when saving a document it errors seemingly because there is still lots of `docPath`s in the `documentPaneStore.ts` and `documentPaneActions.ts` that I have not dealt with. Could you please do that?


desktop/src/main/document/index.ts
```diff
@@ -1,63 +1,42 @@
...
-  delete documentRefs[filePath]
-}
-
 export const registerDocumentHandlers = () => {
   ipcMain.handle(
     CHANNEL.GET_DOCUMENT_CONTENTS,
     async (ipcEvent, filePath: string) => {
-      getDocument(filePath)
-      documents.subscribe(filePath, ['modified'], (_topic, docEvent) => {
+      let documentId = documents.open(filePath).id
```
nokome, line 9: Having two `documents.open(filePath)` in this file doesn't seem right. Instead, I imagine that the `AppDocumentPane` would open the document (when user clicks on a file in the sidebar) and then pass the `documentId` rather than the `filePath` to the channel handlers.

desktop/src/renderer/components/app-document/app-document-editor/app-document-editor.tsx
```diff
@@ -14,35 +14,35 @@ export class AppDocumentEditor {
 
   private editorRef: HTMLStencilaEditorElement | null = null
 
-  @Prop() filePath: string
+  @Prop() documentId: string
 
   private file?: File
```
nokome, line 19: I'm uneasy about the editor retaining a reference to a `File` (presumably the `File` that it was created for). For a new document, created in the app, there will be no `File` yet, if files are renamed then this could become stale. I see that `File` is used in `fileFormatToLanguage` but I suggest that it would be better to have a `private document: Document` instead since that has a `format` property (optionally for new documents as well).

node/src/types.ts
```diff
@@ -4,20 +4,28 @@
...
    */
   path: string
+  /**
+   * Whether or not the document's file is in the temporary directory.
+   */
+  temporary: boolean
+  /**
+   * The synchronization status of the document. This is orthogonal to `temporary` because a document's `content` can be synced or un-synced with the file system regardless of whether or not its `path` is temporary..
+   */
+  status: 'synced' | 'unwritten' | 'unread' | 'deleted'
```
nokome, line 22: @alex-ketch After a little more thought I realized that both `temporary` and `status` are indeed necessary. However, I have removed `status: new` as that was redundant.

PR #981, opened 2021-06-08
feat(Custom errors): Define custom error types and publish them under the "errors" topic

nokome:
By passing up details of the error we are able to suggest remedial action by the user (e.g. a plugin that needs to be installed). For the Node.js bindings, the usual error message string is thrown for these custom errors, but it is also possible to subscribe to the "errors" pubsub topic to be able to provide notifications to  the user.

See `node/src/errors.test.ts` for example usage and `node/src/types.ts` for the `Error` type(s).


rust/src/errors.rs
```diff
@@ -0,0 +1,59 @@
...
+/// Where possible functions should return one of these errors to provide greater
+/// context to the user, in particular regarding actions that can be taken to
+/// resolve the error.
+#[derive(Error, Debug, JsonSchema, Serialize)]
+#[serde(tag = "type")]
+#[schemars(deny_unknown_fields)]
+pub enum Error {
+    /// The user attempted to call a method that is not implemented internally
+    /// and so must be delegated to a plugin. However, none of the registered
+    /// plugins implement this method. It may be that the use needs to do
```
alex-ketch, line 20:
```suggestion
    /// plugins implement this method. It may be that the user needs to do
```

alex-ketch (approved):
✨

PR #993, opened 2021-06-15
Refactor file and document format properties

nokome:
- [x] New `formats` module and rename `DocumentFormt` to `Format`
- [x] Use `Format` type as property for `File`, `Document` and `DocumentEvent`
- [x] Remove unused `File` properties `sha1` and `mime_type` (may be added back to `Format` if needed).
- [x] Update desktop for new typing
- [x] Update Node.js tests
- [x]  Add `preview: bool` to `Format`.


node/src/types.ts
```diff
@@ -33,22 +53,11 @@ export interface Document {
...
-    /**
-     * Whether or not the format should be considered binary e.g. not to be displayed in a text / code editor
-     */
-    binary: boolean
-    /**
-     * The type of `CreativeWork` that this format is expected to be. This will be `None` for data serialization formats such as JSON or YAML which have no expected type (the actual type is embedded in the data).
-     */
-    type?: string
-  }
+  format: Format
```
alex-ketch, line 60: Nothing major, but want to make sure we're on the same page.
I thought we had decided on keeping the formats as a simple string, and having a lookup function mapping it to the various attributes?

node/src/types.ts
```diff
@@ -480,12 +461,16 @@ export type Method =
   | 'build'
   | 'execute'
 
-export const DOCUMENT_FORMATS: Record<string, DocumentFormat> = {
+export const FORMATS: Record<string, Format> = {
```
alex-ketch, line 464: Should these have the `preview: boolean` attributes?

alex-ketch (commented):
Some minor points I could use clarification on, but looks great otherwise 👍

node/src/types.ts
nokome: Yes, we decided to keep formats a simple string and do the lookup. But then we discussed that for the files browser in particular this would require extra expense to do the lookup for something that will change very infrequently. So I thought we reverted the first decision and decided to go with an object instead of a string.

node/src/types.ts
nokome: Yes, it seems I may have not pushed that commit :/

node/src/types.ts
alex-ketch: ah, pardon the confusion. I meant to say that there will be overhead, but nothing to concern ourselves about—meaning we should use the strings 🙃 
No worries either way

nokome: No worries. It feels like using `Format` instead of `strings` is less brittle (i.e. less code, the lookup only gets done in one place) so lets stick with that.

PR #995, opened 2021-06-16
Project updates

nokome:
Allows for runtime updates to project properties e.g. `theme`, `main_path`.

- [x] Refactor file watching into `ProjectHandler` (analogous to`DocumentHandler`)
- [x] Improve termination of file watching threads when  a `Project` (or `Document`) is closed
- [x] Update properties of the project when the `project.json` file is created, removed or modified
- [x] Publish `ProjectEvent`s (note that `projects.open` in the Node.js bindings now takes two subscriber functions, one for `ProjectEvents`, the other for `FileEvents`)
- [x] support `write` with new property values e.g. `theme`, `main`


node/src/projects.ts
```diff
@@ -30,19 +30,36 @@ export function list(): string[] {
...
- *                   events for the project
+ * @param projectEventsSubscriber A subscriber function that will receive `ProjectEvent`s
+ * @param fileEventsSubscriber A subscriber function that will receive `FileEvent`s
  * @return A project
  */
 export function open(
   folder: string,
-  subscriber?: (topic: string, event: unknown) => unknown
+  projectEventsSubscriber?: Subscriber,
+  fileEventsSubscriber?: Subscriber
```
alex-ketch, line 40: What you think about making the `subscriber` argument an object, leaving room for future growth?

nokome: Actually, IMO we should be consistent with `documents` and have separate `subscribe` and `unsubscribe` functions (we also provide these functions for `errors`). Apart from consistency it gives us the flexibility for future growth and perhaps most importantly will allow porting to a Websocket JSON-RPC approach (where we can't send callbacks but we can still subscribe to particular project events and get notifications).

nokome:
@alex-ketch I'm going to merge this now as it includes a bug fix for the document-preview-not-updating bug that I mentioned (caused by not registering `html` as a `Format` in an easrlier PR). Also has improvements to handling of `config` (saving it back to disk etc).

PR #1037, opened 2021-07-14
fix(Documents): Add `previewable` property to indicate whether preview panel should be opened by default

nokome:


alex-ketch (approved):
💯

PR #1057, opened 2021-07-28
feat(Documents): Allow alteration of path and format of a document

nokome:
Please see doc comments and tests for usage. Feel free to merge if suits needs.


node/src/documents.test.ts
```diff
@@ -33,11 +34,51 @@ test('create', async () => {
...
+      format: expect.objectContaining({
+        name: "md"
+      }),
+      status: 'synced'
+    }),
+  )
+
+  expect(alter(doc.id, "myfile.tex")).toEqual(
+    expect.objectContaining({
+      temporary: false,
```
alex-ketch, line 67: @nokome is this the intended behaviour, or is the documentation incorrect? Or third option, I'm confused 🙃 

>  * If a `path` is provided then the document's `name`, `format`, `status` and `temporary`
  * flag will be updated (`temporary` will be set to `true` and `status` to `unwritten`
  * immediately after the call).

nokome: Docs are wrong, test is correct.

PR #1078, opened 2021-08-06
Project graphs

nokome:
Adds project dependency graphs including semantic analysis of JavaScript/TypeScript, Python and R via `tree-sitter`.


nokome:
@alex-ketch Now this is passing tests, I'm going to merge it. 

In the Node.js bindings you can get the `project.graph` as a `Graph` object (which will then be available via the `ProjectEvent` whenever the project is updated). Alternatively, you can get  a Graphviz dot representation of the graph by explicitly calling: `projects.graph("dot")`.

PR #1093, opened 2021-08-18
Document patches

nokome:
Adds patch operations for diffing and merging documents.

I'm going to park this for a while. Some things still to do:

- [ ] finish adding handling for all `InlineContent` and `BlockContent`
- [ ] how to handle `id`s?
- [ ] merging: transform operations (shift keys based on other ops) and handle conflicts
- [ ] consider adding a `Copy` operation for use with `Vec` (given use of hashes should be fast enough?)
- [ ] review `String` differencing: could be made simpler ala `Vec` diff
- [ ] review macros for error handling e.g. `invalid_keys!` Should these be `attempt`s, as they are or should `apply`return a `Result`?


nokome:
After discussion with @alex-ketch we agreed to merge this now because stuff in here will be able to be used for current work on reactivity in the (also soon to be merged) PR #1121. Before merging:

- [ ] Rename `keys` to `address` for clarity (don't use `path` because of potential confusion with file paths
- [ ] Create issues for outstanding topics in checkboxes in the top comment of this PR

PR #1094, opened 2021-08-18
Project graph refinements

nokome:
@alex-ketch This makes the changes to the JSON serialization of project graphs that we discussed except for making paths relative to the project (which I didn't have time to complete). Feel free to merge and work off this.

Rather than add another id to nodes, I have added an `index` property. This is a stable integer that should not change as nodes are added and removed and corresponds to the integers in the `from` and `to` properties of edges.

The `index` is the internal, stable identifier used by nodes in Rust, so it seemed better to use that than to create a new string based one. Note though that it is an integer id that initially corresponds to the position of the node in the `nodes` list but if nodes are removed, may not after updates.

```json
    {
      "type": "File",
      "path": "/home/nokome/stencila/repos/stencila/fixtures/projects/daggy/article.xml.media/fig1-v1.jpg",
      "index": 33
    },
    {
      "type": "File",
      "path": "/home/nokome/stencila/repos/stencila/fixtures/projects/daggy/article.json",
      "index": 34
    }
  ],
  "edges": [
    {
      "from": 0,
      "to": 1,
      "relation": {
        "type": "Assign",
        "range": [
          0,
          0,
          0,
          0
        ]
      }
    },
    {
      "from": 2,
      "to": 3,
      "relation": {
        "type": "Assign",
        "range": [
          0,
          0,
          0,
          0
        ]
      }
    },
```

### DIFF
--- a/node/src/types.ts
+++ b/node/src/types.ts
@@ -305,9 +305,9 @@ export type Resource =
        */
       path: string
       /**
-       * The address of the node
+       * The id of the node with the document
        */
-      address: string
+      id: string
       /**
        * The type of node e.g. `Parameter`, `CodeChunk`
        */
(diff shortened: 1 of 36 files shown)


nokome:
I've now added a `GraphEvent` as discussed. See `node/src/projects.test.ts` for an example of usage. Also, `path`s are now relative to the project root.

PR #1117, opened 2021-08-30
Improvements to encoding and decoding for various formats

nokome:
This PR includes several improvements to encoding and decoding including:

- implementing RPNGs
- using RPNGs in Pandoc encoding
- implementing R Markdown


nokome:
One of the issues with this PR is that builds are stalling on MacOS and Windows. I suspect that is due to a mutex lock when installing the `chrome` binaries for use for RPNG generation.

PR #1121, opened 2021-09-01
Web client and server

nokome:
This PR adds a TypeScript based WebSocket client to the `web` module and finishes the Rust WebSocket server to receive/send messages to/from that client.

Things done so far:

- [x] Do not allow server to be run as root but with option to override
- [x] Separate endpoint for WebSocket connections `/~ws`
- [x] Check that login, cookies etc work with WebSockets 
- [x] Write up development workflow for connecting to server and serving Javascript: 

image

Some things to do:

- [ ] Display folder listings (currently 404s on paths such as `/`; this will make it easier to navigate around test fixtures etc)
- [x] Implement WebSocket API client
- [x] Implement WebSocket API server (currently just echos)


nokome:
@alex-ketch I've done a bunch of things in `web` (and `rust` to support that):

### Tests
- end-to-end tests using Jest (I realized we could just use the `jsdom` environment)
- tests run on CI (although for some reason this PR does not seem to get run on each push?)

### Client -> Server events

- a revised `main` function which lazily creates the WebSocket `client` and  a `session` when it needs to, in response to user initiated events (see the comments in `index.ts`)

image

- WebComponents could either emit such `CustomEvents` or, alternatively, we can attach listener functions to the components that do the same thing as the event listeners in `main` (i.e. make a RPC request to the server)

### Server -> Client events

- the default subscription handlers emit a `CustomEvent` e.g.

```ts
/**
 * Default handler for session events
 *
 * Dispatches a `CustomEvent` with the type of the event,
 * prefixed with "session:" e.g. "session:heartbeat".
 */
function defaultHandler(event: SessionEvent) {
  window.dispatchEvent(
    new CustomEvent(`session:${event.type}`.toLowerCase(), { detail: event })
  )
}
```
- WebComponents could either listen for those `CustomEvents`, or alternatively, we can write other handlers that call methods of the component to update them

Note: While the session workflow is quite complete, a couple of the document functions are just stubbed out on the sever and don't actually do anything.

PR #1149, opened 2021-09-13
Node execution events & misc refinements

alex-ketch:
- [x] Remove mention of Plugins from desktop client
- [x] Standardize node & document ID naming for custom events
- [x] Remove Graph update listeners when closing Graph view for improved performance
- [ ] Web client: Implement node update logic functions for document subscription handlers
- [ ] Add tracing debug calls to project graph unsubscribe calls
- [ ] Update event names and payload structures to match server output (probably better as a separate issue)


web/src/documents.ts
```diff
@@ -76,7 +76,7 @@ export async function close(
   documentId: DocumentId
 ): Promise<Document> {
   return client.call('documents.close', {
-    id: documentId,
+    documentId,
```
nokome, line 79: These changes to JSON-RPC parameter names need corresponding changes to the names on the server-side e.g. https://github.com/stencila/stencila/blob/80e13023befcfdc6bb579f2dd4388a807fa28ca4/rust/src/rpc.rs#L283

web/src/index.ts
```diff
@@ -73,16 +73,17 @@ export const main = (
   // a the "run button" on a `CodeChunk` (to execute it without changing it)
   window.addEventListener('document:node:execute', async (e) => {
```
nokome, line 74:
```suggestion
  window.addEventListener('document:execute', async (e) => {
```

web/src/index.ts
```diff
@@ -73,16 +73,17 @@ export const main = (
...
     const [client, document] = await startup()
-    const { id, value } = (e as CustomEvent<documents.NodeExecute>).detail
-    await documents.execute(client, document.id, id, value)
+    const { nodeId, value } = (e as CustomEvent<documents.NodeExecute>).detail
+    await documents.execute(client, document.id, nodeId, value)
   })
 
   // Listen for a `document:node:changed` custom event emitted from within browser window
   // e.g. user changes the code of a `CodeChunk`, or slides a numeric `Parameter`
   window.addEventListener('document:node:changed', async (e) => {
```
nokome, line 82:
```suggestion
  window.addEventListener('document:patched', async (e) => {
```

nokome (commented):
Thanks @alex-ketch , LGTM. I've made some minor suggestions. I will make those changes at the end of my day if you have not already.

nokome:
I've fixed the server-side PRC parameter names to be consistent with what is used in `web` - tests now pass. I have not changed the naming of the events, and propose to do that in a forthcoming PR focused on that. Given this has fixes for other things (like plugins UI) I will merge and remaining tasks can be carried over to new PR.

PR #1155, opened 2021-09-15
DOM patch operations

nokome:
This PR creates a new type `DomOperation` that is the browser DOM equivalent of a patch `Operation`. A `DomOperation` is similar to an `Operation` but instead of having a `value: Box<dyn Any>` property has a `html: String` property. The aim is to allow an `Operation` to be transformed into a `DomOperation` which can be applied to the DOM by the `web` client.

This requires revisions to generated HTML in `encode::html` such that an `Address` has the same meaning in a `DomOperation` as it does in the `Operation` it was derived from.

In addition to the syncing of changes from server to client, this approach should also allow `DomOperation`s to be used to create visualisations of differences between documents.


nokome:
@alex-ketch I've added a `web/src/patches.ts` module which takes a `DomPatch` (a set of `DomOperation`s) and applies them to the, ummmm,.. DOM. It is very similar to the code in the Rust `patches` module (which operates on the actual nodes).

Overall, I'm happy with the direction this is headed. It is different from what we originally envisaged (using `id`s on all the element) but that is necessary because we can't reflect all document changes using just id's e.g. adding a new paragraph, or moving a paragraph. I explored alternative directions, including trying to take the original direction, but this is by far the simplest: in particular having a direct mapping from `Operation` to `DomOperation` feels like a good approach. 

It also allows us to generate GitHub style HTML representations of differences between documents (i.e. a `viewPatch` function that would add CSS classes to the DOM to visualize the patches `Add`, `Remove`, `Move` etc operations).

I think the next step is to start generating some `DocumentEvent::Patched` events from the server (in Rust calculate the diff when a document changes and publishing those events, along side and eventually instead of the `encoded::html` topic channel). That will allow us to check that this approach is working as expected. There will need to be some more work done on HTML generation to use the `slot="<property>"` convention which the `resolveAddress` function in `web` uses (instead of the `data-itemprop` convention we used previously).

nokome:
🎉 Now publishing patches when the document changes. This means that in memory changes in the document are synced in the browser using `DomOperation`s rather than whole page reloads. Nice to see operations such as `Move` and `Transform` working:

Peek 2021-09-19 20-05

One small bug discovered so far concerns Unicode emoji. Server calculates operation string indices using Unicode characters, whereas it seems like the browser might use Unicode graphemes for indices. Should be a straightforward fix.

nokome:
I've fixed the issue with Unicode graphemes: now both the Rust server and the JavaScript `web` client use graphemes as the basis for string patch operations. This requires adding dependencies for Unicode segmentation to both `rust` and `web` but provides for far more human-understandable string diffs when more complicated Unicode letters are involved.

web/src/patches/utils.ts
```diff
@@ -0,0 +1,182 @@
...
+ * This should only happen if there (a) the client has missed a `DomPatch`
+ * such that the state of the DOM is out of sync with the server-side document, or
+ * (b) if there is a bug in the following code. Hopefully testing rules out (b).
+ *
+ * Reloads the document to get a new DOM state and then throws an exception for
+ * early exit from the calling function.
+ */
+export function panic(message: string): Error {
+  // TODO reload the document
+  return new Error(message)
```
alex-ketch, line 20:
```suggestion
  location.reload();
```

Don't think it applies in our current use cases, but we might need to be considerate of  `same-origin` policies for.

> The reload may be blocked and a SECURITY_ERROR DOMException thrown. This happens if the origin of the script calling location.reload() differs from the origin of the page that owns the Location object. See Same-origin policy for more information.
> from MDN https://developer.mozilla.org/en-US/docs/Web/API/Location/reload

alex-ketch (commented):
Great work @nokome, it seems like things are stable enough to merge https://github.com/stencila/designa/pull/228 now, no?

I do have a question regarding `slots` though.
My first thought was that these were [WebComponent `slot`s](https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_templates_and_slots#adding_flexibility_with_slots), but that doesn't seem to be the case as far as I could tell.
I suspect it was to avoid usage of a `data-` prefix?

If the above holds, I'd be more comfortable with changing the name to avoid unexpected behaviour with elements placement and signal that these aren't WebComponent slots.
As for the name, not sure I have helpful suggestions :(
`data-slot`, `data-key`?

web/src/patches/utils.ts
nokome: Yeah, I had `location.reload()` earlier but removed it because it broke the tests (`jsdom` does not support it?) but also because I think we can do better than reload the entire page i.e. ask for a new complete HTML version of the document and replace the `slot="root"`

nokome:
Actually I changed the name in Rustland from `Key` to `Slot` specifically because I wanted to use the `slot` attribute and thereby avoid the redundancy of having both `slot` and `data-itemprop` on things like the "text" property of a `CodeChunk`. Although `slot` is intended for use with WebComponents, it is a global attribute, so is valid on any element, whether or not it is a child of a custom element.

Currently we use `data-itemprop`, `itemtype` and `slot`, in different contexts, to identify a property of a `Node` when it is represented as HTML e.g.

```html
<span data-itemprop="givenNames">
     <span itemprop="givenName">Peter</span>
     <span itemprop="givenName">Paul</span>
</span>
```

I propose instead replacing `data-itemprop` (which has always felt awkward) with `slot` and using that attribute as our "this element represents property X" attribute regardless of whether the element for the particular `Node` being represented in HTML is a WebComponent or not:

```html
<span slot="givenNames">
     <span itemprop="givenName">Peter</span>
     <span itemprop="givenName">Paul</span>
</span>
```

I suppose I'm more concerned with consistency in the HTML generation code (always using `slot`, and not having to switch between `slot` and `data-slot` depending upon the node type) than potential confusion when it comes to reading the HTML.

Also, if possible, I'd prefer to use `DomPatches` (i.e. server generated HTML) to update WebComponents, particularly the `outputs` of `CodeChunks` and the `output` of `CodeExpressions` (and therefore would like to use `slot` for everything). That would simplify the API flow to:

client -> `Patch` (i.e. `Replace` operations with JSON `value`) -> server -> `DomPatch` (i.e. `Replace`, `Add` operations with `html`).

This would simplify things on both the server and client because (a) only one type of patch would be emitted by the server, and (b) HTML generation does not need to be duplicated on both server and in WebComponents. I feel these are strong enough benefits, in particular (b) that is is worth pursuing this approach.

nokome:
There is now a lot in this PR, including some non `DomPatch` related improvements. Therefore, I have merged it although not fully complete.

PR #1208, opened 2021-10-19
ProseMirror code node renderer

### LINKED ISSUE #1201: Web: Use CodeMirror rendering `CodeBlock` nodes
The article editor schema includes temporary stand-ins for `CodeFragment` and `CodeBlock` e.g.

https://github.com/stencila/stencila/blob/67e92f209853503adc650b89b63a686bbe9b943c/web/src/components/article/edit/schema.ts#L229-L237

This issue is to do what it says in that comment and replace those with CodeMirror-based embedded editors. The approach described in https://prosemirror.net/examples/codemirror/ has the advantage over using WebComponents for this (?) of being able to move the cursor in/out of the code, selections etc. In addition to be able to edit the `text` the user should also be able to change the `programmingLanguage`.

The approach taken here will be extended to editing `CodeExpression` and `CodeChunk` nodes in the ProseMirror editor.

alex-ketch:
Closes #1201

I think this PR is in a good state and can be merged.

There is one bug that I haven't quite been able to pinpoint and would appreciate your help @nokome.
When copy/pasting `CodeBlock` nodes, the programming language is preserved in the component, and correctly set in the logged patch message.
Upon refreshing the page however, the programming language attribute seems to be lost.

Let me know if you'd like more detailed steps on how to reproduce, or have other feedback on the PR.

Thanks!

### DIFF
--- a/web/src/components/article/edit/convert.ts
+++ b/web/src/components/article/edit/convert.ts
@@ -108,7 +108,7 @@ export function transformProsemirror(value: JsonValue): JsonValue {
       // Ensure `text` and get properties
       object.text = (object.content as [string] | undefined)?.[0] ?? ''
       delete object.content
-      object.programming_language = (object.attrs as JsonObject)
+      object.programmingLanguage = (object.attrs as JsonObject)
         .programmingLanguage as string
       delete object.attrs
       return object
--- a/rust/src/patches/boxes.rs
+++ b/rust/src/patches/boxes.rs
@@ -116,7 +116,7 @@ mod tests {
         };
         let patch = diff(&a, &b);
         assert_json!(patch.ops, [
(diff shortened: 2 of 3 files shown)


web/src/components/article/edit/convert.ts
```diff
@@ -108,8 +108,9 @@ export function transformProsemirror(value: JsonValue): JsonValue {
       // Ensure `text` and get properties
       object.text = (object.content as [string] | undefined)?.[0] ?? ''
       delete object.content
-      object.programmingLanguage = (object.attrs as JsonObject)
+      object.programming_language = (object.attrs as JsonObject)
```
nokome, line 111: I think we need a to apply this renaming from snake_case to camelCase, in one place, across all node types and in both directions. Otherwise, I suspect that this will crop again and again. Not sure where the best place for that is yet but will create an issue.

PR #1257, opened 2021-11-03
Serve previews from web client

### LINKED ISSUE #1254: Desktop: Serve document preview from web client
Currently the Desktop client communicates directly with the rust based client, and requests the open document converted to HTML.
The HTML is then injected directly into the preview panel. The preview panel creates a subscription to document updates, and as edits are made, new versions of the document replace the preview contents.
Any change to the source document causes the entire preview document to be replaced.

This is not only inefficient, but more importantly causes loss of state and duplication of document subscription/update logic across clients.

We now have a [Web client](https://github.com/stencila/stencila/tree/master/web) to solve the above problems, and which is designed to manage incremental updates to the document using patches.

---

Several technical challenges:
- [x] Pass Web client server authentication key to the Desktop client
- [x] Pass document path
  - ~Currently the mock server used to validate approach can only serve items from the `/fixtures` directory~ NVM, as per @nokome's comment below, the constraint is based on the path where `stencila` server is started. 
- [x] Ensure multiple projects and documents can be open in the Desktop client at the same
- [x] Ensure Desktop client remains secure
  - To serve the preview contents via iframe, we need to alter our Content Security Policy to include something along the lines of `frame-src="http://127.0.0.1::9000"`. If left as is, then this opens up loading of anything from that endpoint.
  - To harden the iframe, I've added a `sandbox="allow-scripts"` attribute. `allow-scripts` is necessary for custom WebComponents (CodeChunks, etc.) to load.

alex-ketch:
WIP

Closes #1254


nokome:
@alex-ketch Can you please try this out ASAP so I can potentially make any fixes before I head off on holiday.

I've added a `server.serve` function to the Node.js bindings and done some very limited testing using `make try` in the `node` folder:

```
> let stencila = require('.')
undefined
> stencila.server.serve(process.cwd())
'http://127.0.0.1:9000//home/nokome/stencila/repos/stencila/node?token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzUxMiJ9.eyJleHAiOjE2MzczMDI2NzcsInNjb3BlIjoiL2hvbWUvbm9rb21lL3N0ZW5jaWxhL3JlcG9zL3N0ZW5jaWxhL25vZGUifQ.aX6n1XVxEFYu0x67CF2XcmpBRjn8uDjIaQNsK6jVc9m97GMEmUfNx_wHKX6M9TMESPiT292OO5mvq9oL8_MJ8Q'
```

I suspect you'll want to pass the full path of the project and then parse the returned URL into it's components so the port number, token etc can be used with iframes for individual files. Alternatively, instead of a string, we could return an `object` with `port` and `token` properties but I'm not sure that is warranted given the `URL` API that's available in JS.

Note that `stencila serve` is now `stencila server start` (this name change enable explicit stopping, and restarting of the server when in interactive CLI mode).

alex-ketch:
@nokome I'm sorry but I wasn't able to test this, did some commits fail to get pushed up perchance?
Unfortunately I wasn't able to try and manually define the function myself either :/

nokome:
I've pushed those commits.

Note that currently the token that is returned in the URL has an expiry of 60 seconds. In the future, it might be lower and one-time use enforced. That's because, given that it is in a string, it may be exposed - for example using `ps aux` when a browser window is spawned to open a document from the command line. The temporary-in-the-URL token is superseded by a more secure HTTP-only cookie token.

The upshot of this is that you shouldn't probably actually parse out and store the token as I suggested before but instead call `serve` for each new document you want a preview for and use that directly as the source for the iframe.

PR #1276, opened 2021-11-09
Patches & WebComponents

nokome:
This aims to define a HTML encoding for document nodes that are represented as WebComponents (at present only `CodeChunk` and `CodeExpression`)  that  can:

1. have patches applied to it, including properties such as `programmingLanguage`, `text`, `outputs`, `errors` and `caption`
2. be hydrated as a WebComponent

For `CodeChunk`s the HTML in this PR currently looks like (this is from on a new fixture `fixtures/nodes/article-code.json`)

```html
 <stencila-code-chunk itemtype="http://schema.stenci.la/CodeChunk" itemscope  data-programminglanguage="python">
      <meta itemprop="programmingLanguage" content="python"/>
      <pre data-itemprop="text" slot="text"># Some code</pre>
      <div data-itemprop="outputs" slot="outputs">
        <span itemtype="http://schema.org/Text" itemscope>a string</span>
        <span itemtype="http://schema.stenci.la/Boolean" itemscope>true</span>
        <span itemtype="http://schema.stenci.la/Integer" itemscope>42</span>
        <span itemtype="http://schema.stenci.la/Number" itemscope>1.23</span>
        <code itemtype="http://schema.stenci.la/Array" itemscope>[
  1,
  2,
  3
]</code>
        <table itemtype="http://schema.stenci.la/TableSimple" itemscope>
          <caption>
          </caption>
          <tbody data-itemprop="rows">
            <tr itemtype="http://schema.stenci.la/TableRow" itemscope>
              <td itemtype="http://schema.stenci.la/TableCell" itemscope>
                <span itemtype="http://schema.stenci.la/Integer" itemscope>1</span>
              </td>
              <td itemtype="http://schema.stenci.la/TableCell" itemscope>
                <span itemtype="http://schema.stenci.la/Integer" itemscope>2</span>
              </td>
              <td itemtype="http://schema.stenci.la/TableCell" itemscope>
                <span itemtype="http://schema.stenci.la/Integer" itemscope>3</span>
              </td>
            </tr>
          </tbody>
        </table>
      </div>
      <div data-itemprop="errors" slot="errors">
        <div itemtype="http://schema.stenci.la/CodeError" itemscope>
          <span data-itemprop="errorMessage">Some error message</span>
          <span data-itemprop="errorType">SomeErrorType</span>
          <pre data-itemprop="stackTrace">Several lines
whose exact format
will depend upon the language</pre>
        </div>
        <div itemtype="http://schema.stenci.la/CodeError" itemscope>
          <span data-itemprop="errorMessage">Another error message</span>
          <span data-itemprop="errorType">AnotherErrorType</span>
        </div>
      </div>
</stencila-code-chunk>
```

Some things to note:

- the `web/patches/dom` module will apply patches to `data-<lowecase-property-name>` e.g. `data-programminglanguage` attributes if they are present. We could use a kebab-case (or any other) convention here but using `.toLowerCase()` avoids an extra dependency / code. Prefixing with `data-` seems like the Right Thing To Do.
- the `<meta itemprop="programmingLanguage" content="python"/>` element will not get updated by patches; it is just there for Microdata compatability

I've done some preliminary testing and the patches seem to be getting applied OK but needs a bit more work to support a variety of output types (both initial rendering and generation of patches).


alex-ketch:
> the web/patches/dom module will apply patches to data-<lowecase-property-name> e.g. data-programminglanguage attributes if they are present. We could use a kebab-case (or any other) convention here but using .toLowerCase() avoids an extra dependency / code. Prefixing with data- seems like the Right Thing To Do.

Unfortunately I don't think we can avoid having extra code/logic.
For custom WebComponents it's not necessary to use a `data-` prefix for custom attributes. For example, if you take a look at the [`CodeChunk` component documentation](https://github.com/stencila/designa/tree/master/packages/components/src/components/codeChunk#properties) you can see that when using properties in HTML they are lower-kebab-case.

> 
> ## Properties
> 
> | Property              | Attribute              | Description                                                                                             | Type                                                          | Default     |
> | --------------------- | ---------------------- | ------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------- | ----------- |
> | `autofocus`           | `autofocus`            | Autofocus the editor on page load                                                                       | `boolean`                                                     | `false`     |
> | `codeChunk`           | --                     | Stencila CodeChunk node to render                                                                       | `CodeChunk \| undefined`                                      | `undefined` |
> | `executeHandler`      | --                     | A callback function to be called with the value of the `CodeChunk` node when executing the `CodeChunk`. | `((codeChunk: CodeChunk) => Promise<CodeChunk>) \| undefined` | `undefined` |
> | `isCodeVisible`       | `is-code-visible`      | Whether the code section is visible or not                                                              | `boolean`                                                     | `false`     |
> | `keymap`              | --                     | Custom keyboard shortcuts to pass along to CodeMirror                                                   | `KeyBinding[]`                                                | `[]`        |
> | `programmingLanguage` | `programming-language` | Programming language of the CodeChunk                                                                   | `string \| undefined`                                         | `undefined` |

The first column refers to the JavaScript object property keys, and is the casing which should be used when setting them directly using JS. The second column shows how to set these properties using HTML attributes.

## Slotted content replacement

Since we're using `MutationObserver`s for detecting changes to certain things (e.g. CodeChunk `text` property), for performance reasons it seems best if these elements were removed and replaced wholesale, rather than just updating deeply nested DOM tree.
However we should reevaluate this approach and see if which tradeoff leads to better performance—observing the entire subtree of the slotted content, or replacing and re-rendering all nodes for a slotted child.
Having said that, I'm not sure if we can have a universal approach for all scenarios, as it will likely depend on the type of content.

nokome:
> when using properties in HTML they are lower-kebab-case

No worries, an easy change.

> best if these elements were removed and replaced wholesale, rather than just updating deeply nested DOM tree

The current approach will indeed apply as fine grained patches as possible. But for something like the `text` slot these will not be deeply nested - it's just a change to the `textContent` in the `<pre data-itemprop="text" slot="text">`. For the slots where there is deeper nesting e.g. `outputs` there should be no need to create a `MutationObserver` since the patches will do the updating? In any case let's see if there are performance issues with the current approach before making patches more coarse grained.

nokome:
@alex-ketch Could you please take this over. I have set up the server to default to serving from a symlinked folder in a sibling `designa` repo. Using the `258-support-slotchange-events` branch things are working well but with a few issues with the Web Components illustrated in the below screencast:

1. When the language is patched the `select` is updated but the syntax highlighting is not
2. Changes to the editor content are sent as patches to the server and received by other subscribers and are updated in the slot, however the content of the CodeMirror editor is not updated.
3. When an output is added, if the node originally had no output, then "no output..." message still shows.

Peek 2021-11-11 20-11


Could you also review my TypeScript code in the `web` module. The `resolveEventNode` function feels a little hacky but maybe the best we can do. Bonus points if you implement the approach (2) described in the comment for `onContentChange` :)

nokome:
Code errors are now endoded like this:

```html
    <stencila-code-error itemtype="http://schema.stenci.la/CodeError" itemscope kind="error">
      <span data-itemprop="errorType" slot="errortype">SomeErrorType</span>
      <span data-itemprop="errorMessage" slot="errormessage">Some error message</span>
      <pre data-itemprop="stackTrace" slot="stacktrace">Several lines
whose exact format
will depend upon the language</pre>
    </stencila-code-error>
```

But the current styling does not separate code the error type and the message e.g.

image

nokome:
@alex-ketch  I've add the `web/patches/codemirror` module which has various functions for converting CodeMirror changes to Stencila `Operation`s:

- `stateToOps`: the previous brute-force single `Replace` op
- `diffToOps`: generates a string diff of content pre- and post- the change (slows down responsiveness of editor noticeably)
- `changesToOps`: translates individual CodeMirror changes to `Operation`s but generates invalid ops when there are Unicode characters

So I've left it as using `stateToOps` for now. I suggest we can come back to this at a later stage and find solutions amongst debouncing / workers / recalculating offsets as grapheme indices

nokome:
@alex-ketch I'm going to move onto #1254. I'd like to base that on top of this PR (since it has some quite important fixes / refactorings to HTML renderings). Once I've got the tests to pass, do you have any objections if I merge?

alex-ketch:
> So I've left it as using stateToOps for now. I suggest we can come back to this at a later stage and find solutions amongst debouncing / workers / recalculating offsets as grapheme indices

That sounds reasonable to me :)
I don't have any objections to merging this as is @nokome

Issue #1286, opened 2021-11-15
Patches: not all on-disk changes being applied in HTML previews


nokome:
@alex-ketch noticed that not all on-disk changes he made to a Jupyter Notebook were reflected in the HTML preview.

I have been unable to reproduce those errors. All the changes I made (and saved) were reflected using the following test setup from the root of this repo on `master`:

- `jupyter notebook fixtures/articles/kitchen-sink.ipynb`
- `cargo run open fixtures/articles/kitchen-sink.ipynb`

![Peek 2021-11-16 12-25](https://user-images.githubusercontent.com/1152336/141870291-8afe2c8a-4c5d-4c83-930c-3a192b213cb0.gif)

@alex-ketch Could you please see if you get errors with this setup, and if you do report any errors in the browser console. If this works could you test if the problem occurs when using from within an `<iframe>`?

nokome:
After some playing around with different document, I've noticed that the problem arises some time around when the file watcher generated a `NoticeRemove` and `Remove` event for the file:

```
  2021-11-16T23:14:52.864778Z DEBUG stencila::documents: Event for document 'do-6fvwpfIzTPLOaYZtSk8h' at '/home/nokome/stencila/repos/stencila/elife-61277_interactive.ipynb': NoticeRemove("/home/nokome/stencila/repos/stencila/.~elife-61277_interactive.ipynb")
    at rust/stencila/src/documents.rs:1133
    in stencila::documents::document_watch with , path: "/home/nokome/stencila/repos/stencila/elife-61277_interactive.ipynb"
```

Note the `.~` prefix on the file path (a temporary file produced by Python / Jupyter checkpointing?). In any case it should be ignored and note interfere with the patching as it currently seems to be doing

nokome:
The above observation was a red herring. Instead it seems that when the files has writes in quick succession that it can be momentarily empty - which caused the code to skip sending a patch.

If this issue persists please reopen.

PR #1287, opened 2021-11-16
Fix: Jupyter Notebook / HTML patches

nokome:
This PR aims to fix any issues with rendering / patching (based on on-disk changes) HTML previews of Jupyter Notebooks (`.ipynb`).

Issues confirmed so far:

- patching of `Object` and `Array` code chunk outputs is not supported (this requires changes to the generation of the Rust types for these which in turn is waiting on merging the `schema` repo into this one).

- [x] rendering of Plotly and Altair images is broken (to do with HTML encoding).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4702,6 +4702,7 @@ dependencies = [
  "maplit",
  "mime_guess",
  "minidom",
+ "node-reshape",
  "notify",
  "once_cell",
  "parsers",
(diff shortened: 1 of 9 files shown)


nokome:
This also...

- fixes #1286 (hopefully)
- enables patching of `Cite` nodes
- enable reshaping of documents (e.g. for grouping of `Cite` nodes into a `CiteGroup` for correct formatting - although this is not yet implemented in the Rust port from Encoda).

PR #1319, opened 2021-12-01
Web: Populate list of executable Editor languages from server

alex-ketch:
Note that this currently only works for in Read/View modes, and not in
edit modes. Support for edit mode is WIP.

Right now the [`resolveEventNode`](https://github.com/stencila/stencila/blob/6295798eccc1733945188e0cbe1d21bab89f7465/web/src/documents.ts#L300) throws an error because the elements generated with the `edit` mode don't contain `id` attributes, which this function relies on.
What's our current strategy regarding node resolution, how would you like me to proceed @nokome?


nokome:
I'm on my phone (broadband is down) so I haven't been able to check this myself, but all the executable nodes should get and `id` attribute in the HTML, regardless of the mode (the HTML generation is independent of mode).

nokome:
So on the question of how to proceed, I think we should ascertain why / fix the allocation of ids to those nodes.

nokome:
@alex-ketch I've added the `kernels.available()` function to the node bindings.

nokome:
I'm not seeing any issues with missing ids on `CodeChunk` elements and the list of executable languages seems to be working fine, I can change the language fine.

image

I can confirm the language was changed using `?format=json`:

alex-ketch:
> I'm not seeing any issues with missing ids on CodeChunk elements and the list of executable languages seems to be working fine, I can change the language fine.

@nokome: To confirm, you're seeing IDs present even when in `?mode=edit`?

> The <stencila-code-chunk> component does not seem to respect the original language of the CodeChunk - even before any user changes - but is rather selecting the first one in the list (?).

Took some head scratching but I think I have it narrowed down and working on a fix. First though, were you seeing this in Firefox? Does it also happen in Chrome/alt browser?

PR #1331, opened 2021-12-13
HTML encoding and patches for code outputs

nokome:
This implements HTML encoding and patching for node types that are common outputs for code execution:

- `Array`
- `Object`
- `Datatable`

All of these need some special consideration for both encoding and patching (e.g. a `Datatable` is column-oriented but its `<table>` representation is row-oriented).


nokome:
Although patching of `Datatable`s is not yet complete, I'm going to merge this so that I can work off it for a related PR on `Parameter` encoding and patching.

PR #1333, opened 2021-12-14
Parameters: HTML encoding and patching

nokome:
This implements HTML encoding and patching for `Parameter` and `ValidatorTypes` nodes. As always, the HTML encoding attempts to be as semantically correct as possible. For parameters it is a little tricky because the `type` and attributes of the parameter's validator (e.g. `StringValidator::min_length`) need to be encoded on a `<input>` element (e.g. `<input minlength=...`). 

So it took a bit more time to get this working with patches but it seems to be working well. I recommend experimenting by opening up http://127.0.0.1:9000/fixtures/nodes/parameter-number.json etc and modifying the JSON to see how the attributes of the `<input>` element change. Also try changing the type of the validator e.g. `NumberValidator` to `BooleanValidator`.

See the new `rust/codec/html/src/snapshots` files for the encoding of parameters e.g.

```html
<stencila-parameter data-itemscope="root" itemtype="http://schema.stenci.la/Parameter" itemscope>
  <label data-itemprop="name" slot="name" for="input-id">pars</label>
  <meta itemprop="validator" itemtype="http://schema.stenci.la/NumberValidator" itemscope/>
  <meta itemprop="value" itemtype="http://schema.stenci.la/Integer" itemscope content="50"/>
  <input id="input-id" slot="input" type="number" min="0" max="100" value="50"/>
</stencila-parameter>
```

Note that patches do not get applied to the `<meta>` elements (they are placeholders) and Web Components should react to changes on `slot=name` and `slot=input`, including the `type` of input.


nokome:
I've changed the slot name to `value` and added encoding of `default`:

```html
<stencila-parameter data-itemscope="root" itemtype="http://schema.stenci.la/Parameter" itemscope>
  <label data-itemprop="name" slot="name" for="input-id">par</label>
  <meta itemprop="validator" itemtype="http://schema.stenci.la/StringValidator" itemscope/>
  <meta itemprop="default" itemtype="http://schema.stenci.la/String" itemscope content="defaultstring"/>
  <meta itemprop="value" itemtype="http://schema.stenci.la/String" itemscope content="hello"/>
  <input id="input-id" slot="value" type="text" pattern="[a-z]+" placeholder="defaultstring" value="hello"/>
</stencila-parameter>
```

nokome:
I've added support for representing parameters in Markdown (you may remember a version of this from some time ago). The syntax needs to be review, and possibly modified. But mainly wanted to this now so that we have an easy way of writing and testing documents with parameters.

image

nokome:
I've added support for HTML encoding of `EnumValidator`s

```
<stencila-parameter data-itemscope="root" itemtype="http://schema.stenci.la/Parameter" itemscope>
  <label data-itemprop="name" slot="name" for="input-id">par</label>
  <meta itemprop="validator" itemtype="http://schema.stenci.la/EnumValidator" itemscope/>
  <meta itemprop="default" itemtype="http://schema.stenci.la/String" itemscope content="A"/>
  <meta itemprop="value" itemtype="http://schema.stenci.la/String" itemscope content="B"/>
  <select id="input-id" slot="value">
    <option value="A">A</option>
    <option value="B">B</option>
    <option value="C">C</option>
  </select>
</stencila-parameter>
```

as well as representation in Markdown,

image

alex-ketch:
@nokome I've pushed up the commit we authored together for partially handling the Parameter component events. For this to work however the [corresponding PR in designa](https://github.com/stencila/designa/pull/275) needs to be merged first.
I will do that once that CI checks have passed.

nokome:
I've made various fixes in both `web` and `rust` but there are a couple of things that I think need changing in Web Components.

- [ ] When changing a boolean parameter value the patch value is always `"true"`
- [ ] The minimum value for the `minlength` of a `StringValidator` should be 0

Patches from client to server seem to be working OK now apart from the above. Patches from the server to client are failing to resolve the target DOM node most (all?) of the time. So that needs more work.

On the UX side, I found things a bit "jumpy" and wonder if the affordance for the parameters (the equalizer icon) should always visible (as with code expressions) and the name more slowly reveal (rather than suddenly jump out).

nokome:
Merging this now with view to fix server -> client patching later.

PR #1338, opened 2021-12-21
feat(Web): Attempt to reconnect to server

### LINKED ISSUE #1198: Web: Client should attempt to reconnect
The web client only attempts 4 reconnects before giving up e.g.:

image

We should [configure](https://github.com/elpheria/rpc-websockets/blob/HEAD/API.md#new-websocketaddress-options---client) here:

https://github.com/stencila/stencila/blob/67e92f209853503adc650b89b63a686bbe9b943c/web/src/client.ts#L10

with unlimited retries and a random time between reconnects (to avoid thundering herd if server goes down and then comes back on line) e.g.

```ts
{
   reconnect_interval: 1000 + 4000 * Math.random() # 1-5 seconds
   max_reconnects: 0 # unlimited
}
```

It would be nicer to have an exponential backoff (e.g. using https://www.npmjs.com/package/exponential-backoff) instead but `rpc-websockets` does not allow for a function to be supplied as a reconnect option. Perhaps a PR - if it's easy.

alex-ketch:
Close #1198

This PR deviates slightly from the plan outlined in the original issue.
> {
   reconnect_interval: 1000 + 4000 * Math.random() # 1-5 seconds
   max_reconnects: 0 # unlimited
}

I've adjusted the parameters so that reconnection attempts continue for around 5 to 15 minutes.
Intent here being that if a server is going to respond, it should hopefully do so within this time period. If it doesn't, then there's likely a bigger issue at hand, and we avoid an endless runaway process if a client is left open in a background tab or some similar scenario. This should also hopefully reduce the impact of the "thundering herd" mentioned in the issue description"

If you'd like @nokome, I can also engage the underlying `rpc-websockets` library maintainers to gauge their willingness to accept a PR to support passing a function to the `reconnect_interval`.

### DIFF
--- a/web/src/client.ts
+++ b/web/src/client.ts
@@ -20,7 +20,10 @@ export async function connect(
   if (typeof token === 'string' && token.length > 0)
     connectUrl += `&token=${token}`
 
-  const client = new Client(connectUrl)
+  const client = new Client(connectUrl, {
+    reconnect_interval: 1000 + 3000 * Math.random(), // random interval between 1 and 4 seconds
+    max_reconnects: 300, // attempt to reconnect for ~5-15 minutes
+  })
   return new Promise<Client>((resolve) =>
     client.on('open', () => resolve(client))
   )


nokome (approved):
Thanks @alex-ketch, looks fine to me. I don't think its a priority to submit a PR to the upstream right now.

Issue #1251, opened 2021-11-01
Add `status` property to executable Schema nodes


alex-ketch:
To provide visibility into node execution status, we'd like to add a `status` property to various executable Schema nodes (i.e. `Code` & `Math`) based nodes.
This `status` property will then be used by our WebComponent renderers.

nokome:
This has been done as part of #1344 and with the release of `@stencila/schema@1.30.0`.

PR #1344, opened 2021-12-29
Kernel tasks

nokome:
Improve management and scheduling of kernel tasks.


nokome:
This is a long running branch with several important changes, not just to kernel tasks (although that was what it started off on). So, I'm going to merge it now as a base for doing to some UI development that make use of those changes, noteably https://github.com/stencila/stencila/issues/1251

PR #1346, opened 2022-01-11
Add autorun option for node execution

nokome:
This implements an `autorun` option for node execution. At present the only way this option can be set is by using comment tags `@autorun never` and `@autorun always`. The default is "needed". Please see the comments on `ResourceAutorun` for an explanation of these variants. 

I've made changes to the `fixtures/articles/code-relations.md` file and added a `now()` function to Calc to facilitate manual testing.

@alex-ketch Could you please review (it's all Rust but well commented and fairly straight forward) and base your component work off this.


rust/graph-triples/src/resources.rs
```diff
@@ -88,15 +88,59 @@ impl Resource {
...
+///
+/// - "autorun" means that the resource is automatically executed, without an explicit
+///   user request do so (but in some cases in response to one).
+#[derive(Debug, Clone, Serialize)]
+pub enum ResourceAutorun {
+    /// Never automatically execute the resource.
+    /// Only execute when the user explicitly runs the resource (or its containing resource).
+    ///
+    /// e.g. a user may tag a `CodeBlock` as `@autorun never` if it is long running
+    /// and they want to check the outputs of previous code chunks before proceeding
```
alex-ketch, line 107: Just to confirm, this means that if `CodeChunk a` is marked as `@autorun never`, but is referenced from `CodeChunk b`. Changes to `a` mark both `a` and `b` as stale, and no matter how many times you run `b` it will always be in a stale status.
User would have to manually run `a`. Then `b` would react to the the update and get automatically re-evaluated?

We could explore showing a warning or something for such cases, and offer to re-run stale upstream nodes for the user.

nokome: Yes, your reiteration is correct. It is also a good example of why it is useful to have separate `executeRequired` and `exexuteStatus` properties. The user might run b to check that it is correct, ie has status `Succeeded` but still want to know that `executeRequired` is `DependenciesChanged`.

PR #1348, opened 2022-01-18
fix(HTML encoding): Refinement to encoding of code dependencies

nokome:
Makes several changes to HTML encoding of the `code_dependencies` and `code_dependents` properties of `CodeChunk`s, and the `code_dependencies` property of `CodeExpression`s (they have no dependents):

- Use `<stencila-code-dependencies>` and `<stencila-code-dependency>` custom elements for both properties but keep as separate lists distinguishable by `slot` attribute

- Use `node-kind` instead of `node-type` to avoid conflict with native DOM attribute name

- Add `label` property and use for `CodeChunk.label` and `Parameter.name` e.g.

```html
<stencila-code-dependency node-kind="Parameter" node-id="pa-1" label="c" execute-auto="Needed">
</stencila-code-dependency>
```

### DIFF
--- a/rust/codec-html/src/encode/blocks.rs
+++ b/rust/codec-html/src/encode/blocks.rs
@@ -151,7 +151,7 @@ impl ToHtml for CodeChunk {
         );
 
         let dependents = elem_placeholder(
-            "stencila-code-dependents",
+            "stencila-code-dependencies",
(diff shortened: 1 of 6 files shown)


alex-ketch (approved):
Thanks @nokome, I've been working with the output from this branch today and things seem to be working as intended

PR #1349, opened 2022-01-19
Add cancellation of node execution

nokome:
This adds cancellation of node execution. Nodes that have `executeStatus` equal to `Scheduled` or `Running` (and their `XXXPreviouslyFailed` variants) should be able to be cancelled. Where possible (i.e. the kernel allows) the execution task should be aborted, otherwise the result of the task should be ignored (i.e. the document node updated with the outputs/messages from the task).

@alex-ketch : I've scaffolded this, adding a `documents.cancel` function that has the same call signature as `documents.execute`. On the server, the implementation is currently crude but should emit a `executeStatus = Cancelled` patch.


nokome:
@alex-ketch This now does 2 out of the three intended types of cancellations:

1. Node is `Scheduled` or `ScheduledPreviouslyFailed`: when cancelled goes to previous status i.e. `Suceeded`, `Failed` or `None`
2. Node is `Running` or `RunningPreviouslyFailed` and able to be aborted (kernel can be interuppted) goes to `Cancelled`
3. Node is `Running` or `RunningPreviouslyFailed` and finished (kernel can not be interuppted, but result is ignored) goes to `Cancelled`

1 and 3 are implemented but 2 is not yet (so long running tasks e.g `sleep` will still be waited upon but will show cancelled when they finish).

Also I added a temporary `windows.stencilaCancel` function for testing. Feel free to remove when cancel button is in place.

alex-ketch:
@nokome In working on the components, I realized that we actually don't use an `execute` event in the web client. Instead we're attaching handlers in the `initComponents` method:

https://github.com/stencila/stencila/blob/ef2be8cf8426eb18ffcd74f8cd7fceb42deb0a67/web/src/index.ts#L91-L112

I've added the following events:
- `stencila-code-execute`, and
- `stencila-code-execute-cancel`

This did raise a couple of questions however:
- We use the presence of the `executeHandler` prop method on the component to determine whether to display a ▶️ button or not. Presuming we want to deprecate the `initComponents` method and not rely on attaching handler to components, should we instead see if `[documents.stencilaWebClient.executableLanguages](https://github.com/stencila/stencila/blob/0da157cf2331d39686821631dc0467304c4dee33/web/src/types.ts#L14).length > 0`?
- How to handle `cancel` events coming from downstream nodes. I imagine when a node is schedule to run, downstream dependent nodes also change status to `scheduled`. Will the execution plan resolve/reverse-engineer which task to cancel when a downstream noes is canceled? Should nodes prior to the cancelled node still be executed?

nokome:
Yes, I was aware of the inconsistency of still using the `executeHandler` and wanted to fix that up sometime, so thanks for adding those events. And yes, using `executableLanguages.length` seems like the best way to enable the play button or not. Although I suppose ideally, the enable-ment of play should be done on a node-by-node basis i.e. it should only be enabled if the language for that cell is in `executableLanguages` (and change if the language changes, as it currently does).

Regarding cancellation nodes other than the one that was clicked on. I have been having related thoughts and think that for both `execute` and `execute-cancel` we need to "modes" (for want of better names):

- `ThisOneOnly` - only execute/cancel the node that was clicked
- `ThisOneAndReleated` - execute/cancel the node that was clicked and all it's upstream dependencies and downstream dependents

As as user, I think that I'd want to be able to just run a single cell, for example to test a fix, without having to worry about upstream or downstream cascades. The same is true for cancel e.g. I notice an error in a single chunk while a bunch are running - I just want to cancel that and then try to make the fix while the others are still going.

To expose these two options I do wonder if the play/cancel buttons need to have two alternatives (perhaps dropdown) but acknowledge that adds to clutter. Alternatively (and additionally?) we could use shift+click (as we do for other buttons) to indicate the cascading execute/cancel.

Currently we have an in between strategy which cancels whole stages if any dependencies fail or are cancelled - but I don't think that is ideal and we need something more like the above.

nokome:
I've thought about this some more and suggest that I implement the following API for `execute` and `cancel`. Then we can decide best how to expose all, or a subset to the user. Just to clarify some terminology...When `execute`-ing a document we generation an execution plan that has two parameters - the `nodeId` of the starting node and the order of execution, `Single`, `Topological` or `Appearance`. Each document only has one current execution plan, so a `cancel` request only has one parameter, the `nodeId` because the order is determined by the current plan.

### Execution plans for the entire document

These calls would be made from buttons on the toolbar for the document:

- `execute(null, Topological)`: execute all code nodes in the document in topological order; default
- `execute(null, Appearance)`: execute all code node in the document in appearance order (what we currently have implemented on the Hub).
- `cancel(null)`: cancel the current execution plan regardless of ordering (no node id is specified, everything gets cancelled)
- `restart()`: discard the current kernel space and create a new (empty one)

### Execution plans beginning a specific node

These calls would be made from buttons on the toolbar for a particular node:

- `execute(nodeId, Single)`: execute the specific node only
- `execute(nodeId, Topological)`: execute the specific node, and any upstream dependencies that are stale (unless  they have `executeAuto == Never`) and any downstream dependents (unless they have `executeAuto == Never`); default
- `execute(nodeId, Appearance)`: execute the specific node and all code nodes appearing after it
- `cancel(nodeId)`: cancel execution of the specific node
- `cancel(null)`: cancel the current execution plan (same as for whole document, but by appearing on node, implies cancelling of a plan originating from that node)

alex-ketch:
> * `cancel(null)`: cancel the current execution plan (same as for whole document, but by appearing on node, implies cancelling of a plan originating from that node)

This is the only part of the proposal that feels prone to complicate logic and cause issues. What do you think about something like, assumption being that we want to cancel the remainder of the execution plan that this node belongs to:
- `cancel(nodeId)`: cancel the current execution plan that this node belongs to
- `cancel(nodeId, StopPropagation)`: cancel execution of the specific node

nokome:
Yep agreed 👍🏽

nokome:
I've implemented this now, these temporary functions I used for testing (feel free to remove and replace with event listeners) illustrate the API:

```ts
    // Temporary functions for testing in the console
    // @ts-ignore
    window.stencilaExecute = async (
      nodeId: null | string,
      ordering: 'Single' | 'Appearance' | 'Topological'
    ) => {
      const [client, document] = await startup()
      await documents.execute(client, document.id, nodeId, ordering)
    }
    // @ts-ignore
    window.stencilaCancel = async (
      nodeId: null | string,
      scope: 'Single' | 'All'
    ) => {
      const [client, document] = await startup()
      await documents.cancel(client, document.id, nodeId, scope)
    }
    // @ts-ignore
    window.stencilaRestart = async () => {
      const [client, document] = await startup()
      await documents.restart(client, document.id)
    }
```

nokome:
@alex-ketch I've added `<stencila-document-toolbar>` to the top of the HTML encoding of `Article`. I'll go ahead and merge this now and we can do another round of changes to the web client once https://github.com/stencila/designa/pull/285 is finished.

PR #1360, opened 2022-01-25
feat(Client): Add event listeners for code execution & kernel restart

alex-ketch:
Before merging this PR, we should merge https://github.com/stencila/designa/pull/285 and update the Components package within `rust/stencila` and `web` directories.

### DIFF
--- a/web/src/documents.ts
+++ b/web/src/documents.ts
@@ -12,6 +12,7 @@ import {
   Patch,
 } from '@stencila/stencila'
 import { Client, ClientId } from './client'
+import { KernelId } from './kernels'
 import { JsonValue } from './patches/checks'
 import * as codemirror from './patches/codemirror'
 import { applyPatch } from './patches/dom'
@@ -214,14 +215,19 @@ export async function cancel(
 }
 
 /**
- * Restart the document's kernel space
+ * Restart one, or all, of the kernels in a document's kernel space
+ *
+ * If `kernelId` is not supplied then all kernels in the kernel space
+ * will be restarted.
  */
 export async function restart(
   client: Client,
-  documentId: DocumentId
+  documentId: DocumentId,
+  kernelId?: KernelId
 ): Promise<void> {
   return client.call('documents.restart', {
     documentId,
+    kernelId,
   }) as Promise<void>
 }
 
--- a/web/src/documents.ts
+++ b/web/src/documents.ts
@@ -232,6 +232,30 @@ export async function restart(
 }
 
 /**
+ * Get a list of kernels in a document's kernel space
+ */
+export async function kernels(
+  client: Client,
+  documentId: DocumentId
+): Promise<void> {
+  return client.call('documents.kernels', {
+    documentId,
+  }) as Promise<void>
+}
+
+/**
+ * Get a list of symbols in a document's kernel space
+ */
+export async function symbols(
+  client: Client,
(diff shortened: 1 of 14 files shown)


web/src/index.ts
```diff
@@ -91,47 +94,40 @@ export const main = (
...
+        detail.ordering
+      )
     }
-    // @ts-ignore
-    window.stencilaCancel = async (
-      nodeId: null | string,
-      scope: 'Single' | 'All'
-    ) => {
+
+    window.addEventListener('stencila-code-execute', (e) => {
```
nokome, line 108: @alex-ketch Is there any particular reason these `window.addEventListener` calls are here and not in `documents.listen`? It feels like having them all in one place - whichever is the best place - would make the code more maintainable. If there is a reason listeners for `stencila-code-execute`, `stencila-code-execute-cancel`, `stencila-kernel-restart` need to be here and not in the `documents.listen` could you document that.

alex-ketch: nope, was focussing on refactoring the `index.ts` file and didn't see that the event listeners were grouped in the `documents.ts` file.
Refactoring now

web/src/documents.ts
```diff
@@ -210,14 +215,19 @@ export async function cancel(
 }
 
 /**
- * Restart the document's kernel space
+ * Restart one, or all, of the kernels in a document's kernel space
+ *
+ * If `kernelId` is not supplied then all kernels in the kernel space
+ * will be restarted.
```
nokome, line 221: @alex-ketch This is the first of several changes I'll make here: now all, or a specific, kernel in the document's kernel space will be restarted.

web/src/documents.ts
```diff
@@ -210,14 +215,43 @@ export async function cancel(
...
+ * Get a list of symbols in a document's kernel space
+ */
+export async function symbols(
+  client: Client,
+  documentId: DocumentId
+): Promise<void> {
+  return client.call('documents.symbols', {
+    documentId,
   }) as Promise<void>
 }
```
nokome, line 256: @alex-ketch I've added these two functions one to get a `Record<KernelId, KernelInfo>` and the other to get a `Record<String, SymbolInfo>`. At present there is a TypeScript type for `SymbolInfo` but not for `KernelInfo`. Its Rust is:

```rust

/// An identifier for a kernel
///
/// This is *not* a UUID but rather a id that is unique to a
/// local kernel space. This allows more useful ids to be assigned
/// e.g. `python`, `r` etc.
pub type KernelId = String;

/// The status of a running kernel
#[derive(Debug, PartialEq, Clone, Serialize, Display)]
#[allow(dead_code)]
pub enum KernelStatus {
    Pending,
    Starting,
    Idle,
    Busy,
    Unresponsive,
    Stopping,
    Finished,
    Failed,
    Unknown,
}

/// Information on a running kernel
///
/// Used when displaying information to the user about currently
/// running kernels.
#[derive(Debug, Clone, Serialize)]
pub struct KernelInfo {
    /// The id of the kernel instance
    pub id: KernelId,

    /// The status of the kernel
    pub status: KernelStatus,

    /// The kernel spec
    #[serde(flatten)]
    pub spec: Kernel,

    /// Whether the kernel is interruptable on the current machine
    pub interruptable: bool,

    /// Whether the kernel is forkable on the current machine
    pub forkable: bool,
}
```

PR #1380, opened 2022-02-02
feat(Web): Expose `Document::create` to `web` client

nokome:
This adds the `create` function to the `documents` module of the `web` client. It can be called with content e.g.

```ts
documents.create(
   client,
   '{"type": "Article", "content": [...]}',
   'json'
)
```

The response will be a document, with an `id` that can be used with functions like `documents.subscribe`.


web/src/documents.ts
```diff
@@ -64,6 +64,21 @@ export interface ContentChangeEvent extends CustomEvent {
...
+/**
+ * Create a new document
+ * 
+ * Optionally pass the content, in some format, for the new document.
+ */
+export async function create(
+  client: Client,
+  content?: string,
+  format?: string
+): Promise<Document> {
```
alex-ketch, line 76: I wonder if we should go with the convention of using objects as parameters, make arguments a bit more self documenting, and account for any future API changes.

alex-ketch (approved):
Looks good to me, thanks @nokome ✨

web/src/documents.ts
nokome: Finally replying to this. I'm OK with using object as parameters as long as it does not introduce too much boilerplate.

We have talked about refactoring towards using classes for `Document`, `Session` etc in which case parameters such as `client` and `documentId` will be moved into class properties. So perhaps lets revisit convention for parameters when / if we do that refactoring.

nokome:
I have rebased this branch `document-create` onto `web-client-build` so that we can more easily comment on things in this PR. Can we continue work on `document-create` branch please?

web/src/documents.ts
```diff
@@ -64,6 +64,32 @@ export interface ContentChangeEvent extends CustomEvent {
   detail: ViewUpdate | string
 }
 
+/**
+ * Create a new document
+ *
+ * Optionally pass the content for the new document.
+ * If `content` is not a string and format is 'json' of undefined,
+ * then `content` will be stringify to JSON.
```
nokome, line 72: @alex-ketch Note that I have changed the parameter of this so that the caller does not need to do `JSON.stringify` first.

web/src/documents.e2e.ts
```diff
@@ -55,3 +64,46 @@ test('basic', async () => {
...
+  let document = await open(client, 'fixtures/nodes/creative-work.json')
+
+  let json = await dump(client, document.id, 'json')
+  expect(json).toMatch(/^{"type":"CreativeWork"/)
+
+  let md = await dump(client, document.id, 'md')
+  expect(md).toMatch(/^A fixture that is a creative work/)
+
+  let rpng = await dump(client, document.id, 'rpng', 'cc-1')
+  expect(rpng).toMatch(/^data:image\/png/)
```
nokome, line 108: @alex-ketch This is how to get a base64 encoded RPNG for a node. See other tests above for exampes of dumping the whole document and using other formats.

node/src/types.ts
```diff
@@ -233,6 +238,7 @@ export interface Patch {
    * The id of the actor that generated this patch e.g. a web browser client, or file watcher
    */
   actor?: string
+  address: Slot[]
```
alex-ketch, line 241: @nokome: Is this field optional, or always defined, or defined but can be empty array?

nokome: Optional. When defined could be an empty array but usually not.

PR #1393, opened 2022-02-15
Schema TypeScript generation

nokome:
Generate `type`s instead of `interface`s for each schema type; other tweaks to setup / CI.


alex-ketch (approved):
Thanks for this @nokome 🙏 
I've added [an additional commit](https://github.com/stencila/stencila/pull/1393/commits/08503a580f12900b0be28ddf148a1ae989fc3514) based on our conversation regarding not causing conflicts between internal primitive types.
This follows [TypeScript's guidance](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#general-types), and avoid [problems like this](https://stackoverflow.com/a/60235961/5686014).

If you feel this is a breaking change, or causes issues I haven't foreseen, please feel free to revert and we can extract into separate PR.

schema/src/bindings/typescript.ts
```diff
@@ -40,13 +40,9 @@ export const generateTypeDefinitions = async (): Promise<string> => {
...
 ${autogeneratedHeader('build:ts', path.basename(__filename), '//')}
 
-type Null = null
-type Boolean = boolean
 type Integer = number
-type Number = number
-type String = string
-type Object = Record<string, unknown>
-type Primitives = undefined | null | boolean | number | string | Array<unknown> | Object
+type ObjectType = Record<string | number, unknown>
```
nokome, line 44: Can we make `type ObjectType = Record<string, unknown>`? This is consistent with the definition in Rust:

https://github.com/stencila/stencila/blob/7d3016a413859256ed4c553adc08bdca9a87807f/rust/schema/src/prelude.rs#L66-L69

This of leaving it as it is, is that an object with a number key will be valid in TypeScript but then will fail to deserialize to memory in Rust.

To be consistent with the other primitives, I think I'd prefer to just use `Record<string, unknown>` everywhere instead of introducing `ObjectType`.

Also, to be consistent with Rust, and with the union type for `Node` we should de-pluralize to `Primitive`

schema/src/bindings/typescript.ts
```diff
@@ -89,12 +85,24 @@ export const typesInterface = (schemas: JsonSchema[]): string => {
  * Convert the `title` of a JSON Schema to the name of a Typescript
  * type, including the interfaces and unions generated below.
  */
-export const titleToType = (title: string): string => {
+export const titleToType = (title: string | number): string => {
```
nokome, line 88: Why does the `| number` need to be added here? Is the `title` of a JSON Schema object allowed to be a number.

alex-ketch: Seems to be leftover from a previous ObjectType typing, I've now removed it.

schema/src/bindings/typescript.ts
alex-ketch: I've made the changes.
Couple questions though:
- Should Array inside `Primitive` be `Array<Primitive>` instead of `Array<unknown>`?
- Can we restrict the Object type inside the `Primitive` type to something more meaningful, rather than being compatible with any and all schema nodes?

nokome: Yes ideally, they would be as in Rust: `Array<Primitive>` and `Record<String, Primitive>` but I wasn't sure  / wasn't able (?) to define recursive types in TS. Please do so if possible.

alex-ketch: Done. There's a weird quirk with recursive types in that:

```diff
type Primitive =
  | undefined
  | null
  | boolean
  | number
  | string
  | Array<Primitive>
- | Record<string, Primitive> // Doesn't work
+ | { [key: string]: Primitive } // Works
```

¯\\\_(ツ)\_/¯

nokome: One last suggestion while we're at it...I don't think `undefined` needs to be in `Primitive`

PR #1399, opened 2022-02-24
feat(Web): Expose helper for generating node identifiers

alex-ketch:
Ports the [UID helper from Executa](https://github.com/stencila/executa/blob/51a9dbd3457b5dd805c6ab5490256cc4820cd36a/src/base/uid.ts#L23) and exposes it via the Web package.

The helper is only exposed in the module bundle, and not the various `browser` packaged.
This can easily be changed however.

@nokome, is there a list of conventions for the `family` identifiers beyond these? Specifically wondering about various math nodes, and if we break code down beyond `cc`.

So far I've gathered:
- do: document
- cc: codeChunk (or is it shared for all code?)
- re: request
- cl: client
- su: subscription
- ta: task
- ke: kernel
- se: session


web/src/utils/uid.ts
```diff
@@ -0,0 +1,74 @@
...
+ * - contain information on the type of object that they
+ *   are identifying
+ * - have an extremely low probability of collision
+ *
+ * Generated identifiers have a fixed length of 32 characters made up
+ * of three parts separated by dots:
+ *
+ * - 2 characters in the range `[a-z]` that identifying the "family" of
+ *   identifiers, usually the type of object
+ *   the identifier is for e.g. `rq` = request
```
alex-ketch, line 41: @nokome These docs are copy pasted, and state that `request` is abbreviated as `rq`, but in [Rust we use `re`](https://github.com/stencila/stencila/blob/117367ccc72abaffa8e2b45a22babed0cc5e8567/rust/node-execute/src/messages.rs#L5).
Shall I update the docs to `re`?

nokome: Yes please

nokome (approved):
LGTM, thanks

PR #1465, opened 2022-06-16
solve bug:same transformation, the calculated digest is different

jm-observer:
1. If there is no sorting, multiple computations will have different digest when there are multiple changes between snapshots 
2. This should be _other_, otherwise it won't find the file that is _Change::Added_ when it write to layer


nokome:
Thanks again @jm-observer for those useful fixes. I've merged with some minor refactoring: https://github.com/stencila/stencila/commit/85b52ab33ac77b0f4e883f4528a86a08b44c9322

PR #1473, opened 2022-07-05
Avoid inconsistent digest caused by folders

jm-observer:
If change_set includes a new folder, the generated sha256 digest will be different, even though the folder name is the same(but the origin super path is different).It's kind of hard to describe the problem, how do I describe it to make it clear?

### DIFF
--- a/rust/images/src/change_set.rs
+++ b/rust/images/src/change_set.rs
@@ -7,7 +7,11 @@ use std::{
 
 use oci_spec::image::{Descriptor, DescriptorBuilder, MediaType};
 
-use archive_utils::{flate2, tar, zstd};
+use archive_utils::{
+    flate2,
+    tar::{self, HeaderMode},
+    zstd,
+};
 use common::{
     chrono::Utc,
     eyre::{bail, eyre, Result},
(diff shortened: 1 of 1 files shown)


rust/images/src/change_set.rs
```diff
@@ -154,7 +158,7 @@ impl ChangeSet {
             };
 
             let mut archive = tar::Builder::new(&mut layer_writer);
-
+            archive.mode(HeaderMode::Deterministic);
```
nokome, line 161: Hi @jm-observer, I don't yet understand the issue (I'll try to add some tests to try to understand it). But does this fix the issue? As far as I can tell changing the header mode just changes the meta data https://docs.rs/tar/latest/tar/enum.HeaderMode.html

nokome: Actually, @jm-observer would you mind providing some shell commands, or ideally some Rust tests that show the issue?

jm-observer: for example, a Dockerfile is as follows. It should generate the same layer digest whenever it is built.  But not now, because the metadata of the config folder is different at different build times, resulting in different digest generated 
```
FROM scratch
COPY ./helloworld /config/helloworld
CMD /config/helloworld
```

jm-observer: ok, wait for me to write a test. It's a little bit clearer

jm-observer: hi, @nokome , i write a simple test for this: _https://github.com/jm-observer/images_test.git_

nokome: Thanks you for those test @jm-observer. I have added similar test to the `change_set` module https://github.com/stencila/stencila/commit/879b3aa681a8ecd0d26f17e82f0e65d9d02e0506 and merged the PR. Thanks!

Issue #1590, opened 2023-05-03
Whether multiple inheritance needs to be supported？


huang12zheng:
When I checked the data,
I found that https://schema.org/Dentist has three parents.
> https://schema.org/LocalBusiness, https://schema.org/MedicalBusiness, https://schema.org/MedicalOrganization

So I wonder if schema-gen also requires multiple inheritance
：）

nokome:
Thanks for pointing this out @huang12zheng , I don't think I had come across multiple inheritance before in schema.org types.

Multiple inheritance is not required right now in `schema-gen` because none of the types that are implemented require it. But sure, if someone needed to include, for example, a `Dentist` node in a document, then to fit with schema.org this would probably be best implemented as multiple inheritance.

I don't think that would be too difficult, and would probably start with modifying

https://github.com/stencila/stencila/blob/4d23ba867390ecd1d326faa00db56fd65cca3da7/rust/schema-gen/src/schemas.rs#L155

to be a `Vec<String>` or something and modifying the property inheritance algorithm to iterate over all ancestors.

However, it would add some complexity so it would probably be best deferred until a use case actually requires it. Happy to review a PR, if you have one of those use cases :)

PR #1591, opened 2023-05-03
feat: #1590 supported multiple inheritance

huang12zheng:
1. Modify `pub extends: Option<String>` to `pub extends: Option<Vec<String>>`,
and add `#[serde(default)]` and `#[serde(deserialize_with = "handle_string_or_array")]` to support deserialization.
3. Modify `extend` to support multiple inheritance.
4. Modify `expand` to support multiple inheritance.
5. Add handle_string_or_array to ensure compatibility with previous files.

### DIFF
--- a/rust/schema-gen/src/schemas.rs
+++ b/rust/schema-gen/src/schemas.rs
@@ -152,7 +152,9 @@ pub struct Schema {
     pub status: Option<String>,
 
     /// The title of the schema that this schema extends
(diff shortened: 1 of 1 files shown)


nokome:
Thanks very much for this PR 🙏🏽 . I'll take a look and make any comments as soon as I can.

nokome:
Merged, thanks again.

Issue #1593, opened 2023-05-06
Extended KEYWORDS


huang12zheng:
https://github.com/stencila/stencila/blob/4743827ab952382780bcc1efb0db25dc9f6e211b/rust/schema-gen/src/rust.rs#LL134C1-L134C17

I think I can expand to 
```rs
static KEYWORDS: [&'static str; 52] = [
    "as", "break", "const", "continue", "crate", "else", "enum", "extern", "false", "fn", "for",
    "if", "impl", "in", "let", "loop", "match", "mod", "move", "mut", "pub", "ref", "return",
    "self", "Self", "static", "struct", "super", "trait", "true", "type", "unsafe", "union", "use",
    "where", "while", // STRICT, 2018
    "async", "await", "dyn", // RESERVED, 2015
    "abstract", "become", "box", "do", "final", "macro", "override", "priv", "typeof", "unsized",
    "virtual", "yield", "try",
];

if KEYWORDS.contains(&module.as_str()) {
    format!("r#{module}")
} else {
    module.to_string()
}
```

Look forward to your opinion

PR #1596, opened 2023-05-07
feat: #1593, Extended Rust-KEYWORDS

huang12zheng:
* Simple test
1. Change one yaml file. For example,
Note.yaml -> True.yaml, and `title: Note`->`title: True`
2.  cd `/private/tmp/stencila_patch/rust/schema-gen` && cargo run
3. do a check for change


nokome (approved):
Thanks for this, LGTM.

Issue #1595, opened 2023-05-06
impl `merge` for schema's extends


huang12zheng:
# purpose
merge data from parents

# Example
* correct
```rs
schema::Dentist::extend(
        // extend(
        &mut dentist,
        vec![LocalBusiness {
            ..Default::default()
        }],
    )
    .unwrap();
```

*  wrong
```rs
schema::Dentist::extend(
        // extend(
        &mut dentist,
        vec![Thing { // Thing is not directly parent
            ..Default::default()
        }],
    )
    .unwrap();
```

# impl
* share
```rs
fn object<T>(value: &T) -> Map<String, Value>
where
    T: Serialize,
{
    serde_json::from_value(json!(value)).unwrap_or_else(|_| unreachable!())
}

pub(crate) fn extend<T, I>(origin: &mut T, values: I) -> Result<()>
where
    T: Serialize + for<'de> common::serde::Deserialize<'de>,
    I: IntoIterator,
    I::Item: Serialize,
{
    let origin_map = object(origin);
    let value_maps: Vec<(String, Map<String, Value>)> = values
        .into_iter()
        .map(|value| (check_type(&value), object(&value)))
        .collect();

    // 1. Check keys
    let mut origin_keys: HashSet<_> = origin_map.keys().collect();

    let err_output: Vec<String> = value_maps
        .iter()
        .filter_map(|(type_name, value)| {
            let value_keys: HashSet<_> = value.keys().collect();
            let duplicate_keys = origin_keys
                .intersection(&value_keys)
                .collect::<HashSet<_>>();

            let result = if !duplicate_keys.is_empty() {
                // Duplicate keys found:

                Some(format!("{type_name} with {:?}", duplicate_keys))
            } else {
                None
            };
            origin_keys.extend(value_keys);
            result
        })
        .collect();

    if !err_output.is_empty() {
        common::eyre::bail!("Duplicate keys found: {:?}", err_output);
    }

    // 2. Merge values
    let mut origin_map = origin_map;
    origin_map.extend(value_maps.into_iter().flat_map(|(_, m)| m.into_iter()));

    *origin = serde_json::from_value(Value::Object(origin_map))?;
    Ok(())
}
pub fn check_type<T>(_: &T) -> String {
    std::any::type_name::<T>()
        .split("::")
        .last()
        .expect("Invalid type name")
        .to_string()
}
```

* schema
```rs
pub trait DentistParent: Serialize {}
impl DentistParent for LocalBusiness {}
[...]

Issue #1594, opened 2023-05-06
impl `Into` Traits for schema's extends


huang12zheng:
for example:
*  one
```rs
impl Dentist {
    pub fn new() -> Self {
        Self {
            ..Default::default()
        }
    }
}

impl Into<LocalBusiness> for Dentist {
    fn into(self) -> LocalBusiness {
        serde_json::from_value(serde_json::to_value(self).unwrap()).unwrap()
    }
}
impl Into<MedicalBusiness> for Dentist {
    fn into(self) -> MedicalBusiness {
        serde_json::from_value(serde_json::to_value(self).unwrap()).unwrap()
    }
}
impl Into<MedicalOrganization> for Dentist {
    fn into(self) -> MedicalOrganization {
        serde_json::from_value(serde_json::to_value(self).unwrap()).unwrap()
    }
}
```

* two:
```rs
impl Dentist {
    pub fn into_local_business(self) -> LocalBusiness{
        self.into()
    }
  ...
}
```

Look forward to your opinion

nokome:
This seems like useful functionality to have. I have a couple of suggestions re the implementation:

- It is generally recommended to implement `From` (which gives you `Into` for free) rather than `Into` where possible. https://ricardomartins.cc/2016/08/03/convenient_and_idiomatic_conversions_in_rust

- Using serialization/deserialization to implement this seems hacky and is likely to have poor performance. I think a better approach would be to write a `FromAncestors` derive macro which would be able to generate `From` implementations for each ancestor

I've been doing quite a bit of work with derive macros recently to support format codec (e.g. https://github.com/stencila/stencila/blob/main/rust/codec-markdown-derive/src/lib.rs) so would be happy to get that started on that if you are still interested in having this functionality.

nokome:
Closing, as inactive, and no other requests for this feature.

PR #1598, opened 2023-05-07
feat: #1595 #1594 impl into and merge [parents]

huang12zheng:
base on #1597 

1. artifacts just add two line with proc-macro
2. An example of the output can be found in the comment of proc-macro
3. limit proc-macro ‘s merge only with parent 
4. Remove any of the possible duplicates (without affecting the result)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -519,6 +519,7 @@ dependencies = [
  "json5",
  "maplit",
  "once_cell",
+ "paste",
  "regex",
  "serde",
  "serde_json",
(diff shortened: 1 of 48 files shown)


huang12zheng:
I just found out that schema filters abstract -_-||

huang12zheng:
```
// Node union type
        // Order is important for deserialization correctness and performance since
        // serde attempts to deserialize in the order in the emum. We put primitives
        // first (for fast deserialization for kernel outputs) excecpt for `Object` which is
        // last so it does not "consume" entity types (which are also objects).
        let primitives_set = [
            "Null",
            "Boolean",
            "Integer",
            "UnsignedInteger",
            "Number",
            "String",
            "Array",
        ];
```
https://github.com/stencila/stencila/blob/main/rust/schema-gen/src/schemas.rs#L355-L362
I don't think the types here should be written dead, maybe configured in yaml and then filtered
Maybe `schema.type` has this information

nokome:
Thanks for the PR @huang12zheng. Could you please rebase this branch onto current `main` so it is easier for me to review (given that this currently includes various other changes already accepted).

huang12zheng:
I updated the introduction of this PR @nokome

nokome:
Sorry for not reviewing this earlier. Given my comments in #1594 and #1595, particularly about using derive macros instead on relying on  serde, I'm going to close this PR. If you still need this functionality then please don't hesitate to let me know and I'd be happy to work with you to implement it using that approach.

PR #1648, opened 2023-09-05
chore(docs): review of manual cli install instructions for linux

ignatiusm:
- [x] review auto install script
- [x] review manual install via tarball
- [x] review docker (after amendment by @nokome )

This PR adds one line to `README.md` cli install instructions: Users need to `cd` into the recently extracted directory before moving the binary to `/usr/local/bin`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -48,6 +48,7 @@ To install a specific version, append `-s vX.X.X`. Or, if you'd prefer to do it 
 
 ```console
 tar xvf stencila-*.tar.xz
+cd stencila-*/
 sudo mv -f stencila /usr/local/bin # or wherever you prefer
 ```
 


ignatiusm:
PR is still work in progress. Will update PR after reviewing docker.

nokome:
I've managed to find out how to make `ghcr.io/stencila/stencila` public and that works for me now.

Also, given that previous version of the image were already published to https://hub.docker.com/r/stencila/stencila, I've updated the `release.yml` workflow to also publish to there and updated the README accordingly. Release in progress so may take a few minutes to push up.

nokome:
Thanks for this, only one line, but it's great to have someone check those instructions (and make fixes) - and the Docker usage is improved now too!.

Issue #1791, opened 2023-12-05
feat(Source view): Flesh out editor


nokome:
The current editor is intentionally very bare bones ☠️ . This issue is to do some re-implementation of some of the niceties of the previous CodeMirror editor: https://github.com/stencila/stencila/blob/v1/web/src/components/editors/code-editor/code-editor.ts

That previous editor was a Web Component that was used for (a) editing the entire source of a document (e.g. in Markdown), (b) the code in `CodeChunk` and `CodeBlock` nodes, and (c) the code in `CodeExpression`, `IfBlock`, `ForBlock` etc nodes (single line mode).
Trying to fit these different use cases into a single component was OK, but a little forced. In particular, the (a) use case which is the use case of the `SourceView` is quite different from the others in terms of its UX and how it read/writes from/to the document (the whole document, not just nodes within it). Therefore, I suggest we add this functionality to the existing `SourceView` component and implement another, different component for the other use cases later.

A list of features to add:

- [x] Syntax highlighting for the languages that we currently support for documents
     - [x] JSON
     - [x] JSON5 
     - [x] YAML
     - [x] HTML
     - [x] JATS XML
     - [x] Markdown
 - [x] Line numbers
 - [x] Word wrap
 - [x] History undo/redo
 - [x] Indentation, bracket highlighting, etc..
 - [x] Autocomplete for CodeMirror and third party plugins

nokome:
Closing as all items in list are now done 🙏🏽 .

PR #1793, opened 2023-12-06
Feature/sourceview syntax highlights 1791

mike-parkin:
**details**

Have implemented the following lang support to the source view:
 - json,
 - json5,
 - jats (currently using lang-xml only),
 - html,
 - yaml,
 - markdown
It seems jsonld is also supported by the LanguageSupport for json, so that is a bonus,

Have also implemented the basic linenumber gutters (with active line highlighting) with the default styling.

_nb. the xml and html will currently load unformatted in the editor, and will often not have closing `</p>` tags, I can address this issue before this PR is approved, but it doesn't appear to be a simple fix_

**requirements**

New deps have been added to the `web/` application, so will need to run a fresh `npm install` before testing

**related issues**
#1791

**to test**

run the cargo server and parcel, navigate to one of the examples in the src view and set the format query parameter to one of the above options:
eg: http://127.0.0.1:9000/examples/nodes/note/note?view=source&format=yaml


nokome:
Thanks @mike-parkin, looks good. I've added three commits:

- some formatting slight refactoring, and documentation
- pinned dependencies (it's the general approach we take in this repo)
- fixed the issue with compact XML and HTML

PR #1799, opened 2023-12-07
feat: app and print view

nokome:
This does some refactoring around views and adds a `web/src/apps/main.ts` file which is served if there is no `?view` query parameter specified.

It's also going to add some skeleton theme's just to prove out end-to end functionality for that.


nokome:
@simonwinter This addresses some of the temporary hacks we made on your branch. Some differences to what we discussed:

- instead of `web/src/app/app.ts` I've used `web/src/apps/main.ts` (for consistency of naming of other folders and in case we have other apps e.g. a lightweight one for some use case)

- the web component in that file should be named `<stencila-main-app>` and should take `doc` as an optional `@property`: if that property is provided then "open" that document (pass that id down to the view components)

- to get the the app to load, just don't add the `?view` parameter i.e. it is now the default.

PR #1801, opened 2023-12-07
Feature/sourceview text functions

mike-parkin:
** details **

Add in word wrapping, with a checkbox to toggle on and off.

Add in histroy with default key mappings for undo/redo undo select/redo-select.

Add in the base autocomplete for code mirror, with some keymappings

Add in some more basic editor extensions: bracket highlighting, fold gutter, dropCursor

I have added the indentOnInput() to the extenions, however this soes not seem to have the desired effect, at least from what I was expecting ,we can discuss later.

** related issues ** #1791

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -12765,6 +12765,7 @@
       "version": "2.0.0-alpha.23",
       "license": "Apache-2.0",
       "dependencies": {
+        "@codemirror/autocomplete": "^6.11.1",
         "@codemirror/commands": "^6.3.2",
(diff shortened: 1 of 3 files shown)


nokome:
Thanks for this @mike-parkin , I've made a few minor changes:

- Tweaks to some of the documentation
- `protected async update` to `override update`: because it does not need to be async and because `override` is a nice built in way of documenting why a method exists
- I moved the `render()` method up to below `styles` just to maintain a top-down ordering to rendering functions

PR #1811, opened 2023-12-10
feat(views/source): add in doc search key mappings

mike-parkin:
**details**

add the search key mappings from @codemirror/search to enable the crtl-F

have set the search interface controls to appear at the top.

currently using default options for keyMappings and search config,

The codemirror search options will only open if the editor is in focus, otherwise it will default the browser search. if we want to set this that is always focuses on the I suggest creating this in another issues.

The key mappings and the further config options can be found here: https://codemirror.net/docs/ref/#search

**related issues**: closes#1807

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -888,6 +888,16 @@
         "crelt": "^1.0.5"
       }
     },
+    "node_modules/@codemirror/search": {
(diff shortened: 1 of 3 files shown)


nokome:
Works great, thanks!

PR #1816, opened 2023-12-11
feat(agents): initial iteration of AI agents

nokome:
This is a first iteration of integrating AI agents. The `Agent` trait defines the interface (i.e. the methods that each agent must provide) and `OllamaAgent` and `OpenAIAgent` implement that trait. The `agents` crate aggregates the agents together so that they can be delegated to based on the type of generation (currently text or image) that they support. 

This functionality is exposed in the CLI commands `agents` and `generate`. See the `--help` for those for more details.

The OpenAI agents will only show up if you have an `OPENAI_API_KEY` env var. The Ollama agents will only show up if you have an `olllama` server running.

image


nokome:
Done some refactoring to make the `Agent` API more "multimodal" and to fetch the list of models from the OpenAI API:

image

nokome:
I've added some initial prompt engineering scaffolding:

image

Basic usage right now:

- write a prompt in a plain text file in `rust/src/prompts` split into three sections: description, system and user prompts separated by three asterisks
- specify the prompt as a CLI argument

The intention is that each agent can have a default prompt customized to it's specific purpose. See `rust/agent/src/prompts/insert-block.md` for an example.

nokome:
Some things to do while I'm thinking about it:

- [ ] on the CLI, nest the `agents` and `generate` commands in an `ai` command e.g. `stencila ai agents`
- [ ] provide a REPL mode for `stencila ai generate` to avoid each process restart having to fetch the list of models and we can take advantage of the debug reloading of prompt templates
- [ ] use a `Lazy` static for the list of agents and only update once every x minutes and maybe asynchronously
- [ ] create a router agent that will detect new `Instruction` nodes in the document and route to the agent best suited to deal with it

PR #1826, opened 2023-12-13
(feature): app chrome styling

simonwinter:
### Summary

This is the start of styling the application using `twind` and `lit` components. The aim here was to identify elements that could be styled easily, test my assumptions of how `twind` works with `lit` & attempt to establish some initial baseline styling approaches.

> [!IMPORTANT]
> Please note: this is not a design-based PR. Any design here should be considered placeholder until more work is undertaken there.

### Important files to review

- web/src/apps/selector.ts

#### UI

The app (`web/src/apps/main.ts`) includes a layout change to the header. This will be worked on more in design (ignore _exactly_ what it looks like, the concepts are more important here).

CleanShot 2023-12-13 at 16 16 35

CleanShot 2023-12-13 at 16 23 01

I've moved the `view` and `theme` selectors to be closer to the document, set the print preview as an element separate to them & added a "document name" placeholder for future work.

There is a new `selector` component in `web/src/apps`, which is the bulk of the work.

**Selector**

The selector component adds styling to a `<details>` element. This allows us to have a summary (in our case the label) + a number of list items. One could question the semantic choice here - but I think that can be a future PR.

#### Linting

eslint & prettier have been configured for the `web` directory & will be expanded to include `ts` and `node` in a future PR. The linting for `web` includes extending an eslint plugin for lit, removing semicolons and using single rather than double quotes (though attributes passed to html elements remain using double quotes)

**Notes:**

1. The `Twind` documentation is not explicit concerning the use of `tw` within lit - e.g. to access themes (`tw.theme`). The assumption is that one uses:

   ```ts
   import { tw } from '@twind/core'
   ```

   This simply does not work. Using the `tw` namespace, silently errors and kills the entire component. However, the provided decorator gives access to an instance variable `tw` - so, one can use `this.tw...`. There is no type safety here, so I have created a simple wrapper to extend `LitElement`:

   ```ts
   class TWLitElement extends LitElement {
     protected tw: typeof tw
   }
   ```

   It still needs some work, as this kind of statement is not typesafe: `const red = this.tw.theme('colors.brand.red')` - it returns any.

2. Application of tailwind classes. The selector component breaks down most of the smaller parts of the UI elements into instance methods. Inside these, where I'm using more than a handful of tailwind classes, I use the `apply` function:

   ```ts
   const styles = apply([
      'text-base font-bold',
      'leading-none',
      'select-none',
      'appearance-none ',
      'min-w-fit',
      'py-2 px-4',
      'bg-white',
      ...
   ])
   ```
   
   These can then be used as the class name for the element you're rendering. In general this approach leads to cleaner code & makes it easier to debug tailwind issues. As a rule of thumb, I like to have entries in the array grouped by function (e.g. related to text, or padding) with variants where possible placed at the end of the array. Who knows - there may be an eslint rule that could help with enforcing this! - I don't think it's a big concern to address.
   
3. Importing additional css files seems to have issues in parcel; I get an error similar to the following issue: https://github.com/parcel-bundler/parcel/issues/6310. Looking at the `lit` docs, they don't recommend using an external stylesheet at all - https://lit.dev/docs/components/styles/#external-stylesheet


package.json
```diff
@@ -22,6 +22,7 @@
     "depcheck": "1.4.7",
     "eslint": "8.55.0",
     "eslint-plugin-import": "2.29.0",
+    "eslint-plugin-lit": "^1.10.1",
```
nokome, line 25: I've been pinning all deps (no caret) and then leaning on Renovate to keep them up to date. I think that is the recommended approach for "apps" rather than "libraries". But open to discussion.

web/src/apps/main.ts
```diff
@@ -87,118 +106,102 @@ export class App extends LitElement {
...
-              </option>`,
-          )}
-        </select>
-      </label>
-    `;
+    const clickEvent = (e: Event) => {
+      this.view = (e.target as HTMLButtonElement).dataset[
+        'value'
+      ] as DocumentView
+    }
```
nokome, line 117: I like this approach of local functions for event handlers - keep things located together closely without cluttering the template.

web/src/apps/selector.ts
```diff
@@ -0,0 +1,179 @@
...
+ * A selector that updates some display portion of the UI
+ */
+@customElement('stencila-ui-selector')
+@withTwind()
+export class UISelector extends TWLitElement {
+  /**
+   * Ref to allow us to close the details element when needed.
+   */
+  detailsRef: Ref<HTMLDetailsElement> = createRef()
+  /**
```
nokome, line 27: A matter of personal taste but I find the lack of blank lines between properties a little jarring (maybe just because I'm not used to it!)

web/src/apps/selector.ts
```diff
@@ -0,0 +1,179 @@
...
+      'items-center',
+      'group-hover:text-brand-blue group-hover:border-b-brand-blue',
+    ])
+
+    const hideMarker = css`
+      &::marker {
+        display: none;
+        font-size: 0;
+      }
+    `
```
nokome, line 86: Love this use of `apply` and `css` to keep templates and code in general clean

web/src/apps/selector.ts
```diff
@@ -0,0 +1,179 @@
+import { apply, css, tw } from '@twind/core'
```
nokome, line 1: I think this file should go into a new `ui` (or similar name) folder next to `apps` given that the web components in `nodes` are likely to use these UI components too.

nokome (changes requested):
Thanks Simon. Looks great. I've left some comments, mostly minor. Feel free to merge when you've had a look at those.

package.json
simonwinter: Yeah - that's fair. Will change.

web/src/apps/selector.ts
simonwinter: nah - this ain't right. I think this has happened post eslint update. I hate it as well. Will fix!

web/src/apps/selector.ts
simonwinter: I'm going to add some docs for this I think

web/src/apps/selector.ts
simonwinter: ha! I've done this with the new work I'm looking at today! Will migrate that into this PR. Bear with me a moment!

nokome (approved):
LGTM!

Issue #1829, opened 2023-12-13
feat(agents): add `AnthropicAgent`


nokome:
We'd like to experiment with the [Claude](https://www.anthropic.com/product) model for some of our custom agents because it has a very large context window compared to other models. The [anthropic-rs](https://github.com/abdelhamidbakhta/anthropic-rs) crate looks useful and can be used in a similar way to how we use `ollama-rs` and the `async-openai` crates elsewhere. I suggest using our `agent-openai` crate as a starting point.

- [x] `cd rust && cp -r agent-openai agent-anthropic`
- [x] Change the name of the crate in Cargo.toml
- [x] Remove `async-openai` and add `anthropic-rs` in Cargo.toml
- [x] Do a search and replace of openai -> anthropic (taking into account capitalization) in `src/lib.rs`
- [x] It looks like there is no explicit chat like API so just re-implement `text_to_text` and remove `chat_to_text` and `text_to_image` methods
- [x] Add `agent-anthopic` to the `rust/agents/Cargo.toml` file and call its `list` function in the `agents::list` function
- [x] Check that Claude appears in `cargo run --bin stencila agents` list
- [x] Check that Claude can be used `cargo run --bin stencila generate --agent anthropic/claude-v1` (or whatever it is named)

nokome:
Thanks for the PR. All done, so closing!

PR #1830, opened 2023-12-14
Add anthropic agent

brettc:
Implement #1829

### DIFF
--- a/rust/agent-anthropic/src/lib.rs
+++ b/rust/agent-anthropic/src/lib.rs
@@ -12,7 +12,9 @@ use agent::{
     common::{
         async_trait::async_trait,
         eyre::{bail, Result},
+        eyre,
         serde_json::json,
+
     },
     Agent, AgentIO, GenerateOptions,
 };
@@ -34,7 +36,20 @@ impl AnthropicAgent {
             model,
         }
     }
+
+    /// Map the agent name to the full version name.
+    /// We fix this here to allow reproducible output for now.
+    /// https://docs.anthropic.com/claude/reference/selecting-a-model
+    fn get_full_version(&self) -> Option<&'static str> {
+        match self.model.as_str() {
+            "claude-instant" => Some("claude-instant-1.2"),
+            "claude" => Some("claude-2.1"),
+            // Add more mappings here
+            _ => None,
+        }
+    }
 }
+
 
 #[async_trait]
 impl Agent for AnthropicAgent {
@@ -56,25 +71,31 @@ impl Agent for AnthropicAgent {
     }
 
     async fn text_to_text(&self, instruction: &str, options: &GenerateOptions) -> Result<String> {
-
-        let (system_prompt, user_prompt) = self.render_prompt(  
+        let (system_prompt, user_prompt) = self.render_prompt(
             &options.prompt_name,  
             json!({  
                 "user_instruction": instruction  
             }),  
         )?;  
         // Build from configuration.
-        // TODO: We need to use the configuration here.
-        // Currently just a cut/paste from the example.
+        // TODO: We need to add the following which are not in the CompleteRequest (maybe by PR).
+        // https://docs.anthropic.com/claude/reference/complete_post
+        // temperature
+        // top_k
+        // top_p
         let cfg = AnthropicConfig::new()?;
         let client = Client::try_from(cfg)?;
+        let full_version = self.get_full_version().ok_or_else(|| eyre::eyre!("Unknown model"))?.to_string();
 
         let complete_request = CompleteRequestBuilder::default()
-            .prompt(format!("{HUMAN_PROMPT}{instruction}{AI_PROMPT}"))
+            // The .._PROMPT values have embedded carriage returns.
+            // System prompts in Claude are just put before the HUMAN_PROMPT.
+            // https://docs.anthropic.com/claude/docs/how-to-use-system-prompts
+            .prompt(format!("{system_prompt}{HUMAN_PROMPT}{user_prompt}{AI_PROMPT}"))
             // What is claude-instant-1??
-            .model("claude-instant-1".to_string())
-            .max_tokens_to_sample(256usize)
-            .stream_response(false)
+            .model(full_version.clone())
+            // Not sure the best way to do this, but 256 is the default.
+            .max_tokens_to_sample(options.max_tokens.unwrap_or(256) as usize)
             .stop_sequences(vec![HUMAN_PROMPT.to_string()])
             .build()?;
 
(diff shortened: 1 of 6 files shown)


rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+
+use agent::{
+    common::{
+        async_trait::async_trait,
+        eyre::{bail, Result},
+    },
+    Agent, AgentIO, GenerateOptions,
+};
+
+/// An agent running on OpenAI
```
nokome, line 19:
```suggestion
/// An agent running on Anthropic
```

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+///
+/// The environment variable OPENAI_API_KEY must be set to use these agents.
```
nokome, line 21:
```suggestion
/// The environment variable ANTHROPIC_API_KEY must be set to use these agents.
```

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+pub struct AnthropicAgent {
+    /// The OpenAI name for a model including any tag e.g. "llama2:13b"
```
nokome, line 23:
```suggestion
    /// The Anthropic name for a model including any tag e.g. "claude-v1"
```

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    ///
+    /// Used as the required `model` parameter in each request to `POST /api/generate`
```
nokome, line 25: Delete this "Used as..." sentence - it's a hang over from elsewhere.

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+    /// (along with `prompt`).
+    model: String,
+
+}
+
+impl AnthropicAgent {
+    /// Create a OpenAI-based agent
```
nokome, line 32:
```suggestion
    /// Create a Anthropic agent
```

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+            model,
+        }
+    }
+   fn supported_inputs(&self) -> &[AgentIO] {
+        &[AgentIO::Text]
+    }
+
+    fn supported_outputs(&self) -> &[AgentIO] {
+        &[AgentIO::Text]
+    }
```
nokome, line 44: Delete these - they are `Agent` trait methods so must be implemented in `impl Agent for AnthropicAgent`

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    // fn supported_inputs(&self) -> &[AgentIO] {
+    //     &[AgentIO::Text]
+    // }
+
+    // fn supported_outputs(&self) -> &[AgentIO] {
+    //     &[AgentIO::Text]
+    // }
+
```
nokome, line 65: Re-enable these

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+        // Build from configuration.
+        // TODO: We need to use the configuration here.
+        // Currently just a cut/paste from the example.
+        let cfg = AnthropicConfig::new()?;
+        let client = Client::try_from(cfg)?;
+
+        let complete_request = CompleteRequestBuilder::default()
+            .prompt(format!("{HUMAN_PROMPT}{instruction}{AI_PROMPT}"))
+            // What is claude-instant-1??
+            .model("claude-instant-1".to_string())
```
nokome, line 76:
```suggestion
            .model(self.model.clone())
```

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+
+    // fn supported_inputs(&self) -> &[AgentIO] {
+    //     &[AgentIO::Text]
+    // }
+
+    // fn supported_outputs(&self) -> &[AgentIO] {
+    //     &[AgentIO::Text]
+    // }
+
+    async fn text_to_text(&self, instruction: &str, options: &GenerateOptions) -> Result<String> {
```
nokome, line 66: At the top of this function you should render the prompt template with the instruction. This is how we pass context variables into the custom prompts. See `agent-openai` crate.

```rust
        let (system_prompt, user_prompt) = self.render_prompt(
            &options.prompt_name,
            json!({
                "user_instruction": instruction
            }),
        )?;
```

rust/agents/src/lib.rs
```diff
@@ -40,6 +40,11 @@ pub async fn list() -> Vec<Arc<dyn Agent>> {
         Err(error) => tracing::debug!("While listing Ollama agents: {error}"),
     }
 
+    match agent_anthropic::list().await {
+        Ok(mut agents) => list.append(&mut agents),
+        Err(error) => tracing::debug!("While listing Anthropic agents: {error}"),
+    }
```
nokome, line 46: See the note earlier on in this function - we need a better long term solution perhaps, but for now I think Anthropic should go before ollama.

rust/agents/Cargo.toml
```diff
@@ -8,4 +8,5 @@ agent = { path = "../agent" }
 agent-custom = { path = "../agent-custom" }
 agent-ollama = { path = "../agent-ollama" }
 agent-openai = { path = "../agent-openai" }
+agent-anthropic = { path = "../agent-anthropic" }
```
nokome, line 11: Generally I try to maintain alphabetic order here.

nokome (changes requested):
Thanks for this! Looks good but needs some changes particularly the prompt rendering and model name suggestions. The latter may even fix the API issue

rust/agent-anthropic/src/lib.rs
```diff
@@ -0,0 +1,98 @@
...
+        // TODO: We need to use the configuration here.
+        // Currently just a cut/paste from the example.
+        let cfg = AnthropicConfig::new()?;
+        let client = Client::try_from(cfg)?;
+
+        let complete_request = CompleteRequestBuilder::default()
+            .prompt(format!("{HUMAN_PROMPT}{instruction}{AI_PROMPT}"))
+            // What is claude-instant-1??
+            .model("claude-instant-1".to_string())
+            .max_tokens_to_sample(256usize)
```
nokome, line 77: This should really come from options, rather than be fixed. One option would be to use a fallback e.g

```suggestion
            .max_tokens_to_sample(options.max_tokens.unwrap_or(256) as usize)
```

Other options like `temperature` and `seed` (if available) should also be set or log a warning if they can't be set

rust/agent-anthropic/src/lib.rs
brettc: This now gets the full version, depending on which claude model we want.

rust/agent-anthropic/src/lib.rs
brettc: * The max tokens is done. 
* The other options cannot be set, as they are not included in the rust crate. We could do a simple PR to add them (and fork for now, as we already need to use the git repo directly)
* I haven't dealt with logging warnings yet.

PR #1837, opened 2023-12-14
feat(prompts): add the document and any specific node to context

nokome:
This adds makes more context available to the the prompt rendering: the document and any specific node (e.g. for modify instructions e.g. "fix this paragraph"). With the REPL you can specify (at startup and in-session) the path for both document and node and I have created some examples.

The `echo` prompt is a good way to understand how these can be rendered in the prompt:

image

I chose HTML as the default format for the rendering as I thought that would be most semantically understood by the AI model. But this is configurable in the options.

@hlm628 and @brettc could you look at the diffs in this PR to understand what these changes allow in terms of prompt engineering and then add to the `prompts/README.md` file accordingly.


nokome:
I've gone ahead and merge this (with substantial refactoring) so closing now.

PR #1844, opened 2023-12-17
feat(split-view): creation of ui to manage code/editor split view

### LINKED ISSUE #1828: feat(design): app chrome for split view
We need to alter the split view to:

- [x] have vertical side-by-side view
- [x] allow the 2 panes to be resized via a draggable control
~~- [ ] update the chrome surrounding the code editor (e.g. the language selector)~~

simonwinter:
### Summary

This PR introduces a split view to allow users to view both the code editor and preview side-by-side. This introduces a new `split-drag` ui element which provides a `left` and `right` slot for consumers to add content they'd like to see split.

> [!Important]
> This PR does not include any design.

### Important files to review

- web/src/views/split.ts
- web/src/ui/split-drag/index.ts

### New features

Includes a new component `web/src/ui/split-drag/index.ts`. The component provides:

- A `left` and `right` slot
- Depending on viewport size, displays the 2 slots in either a horizontal or vertical split
- The split can be reset by clicking on the drag control

The split is provided by the [split.js](https://split.js.org/) library.

https://github.com/stencila/stencila/assets/1104537/d54f27ee-d892-4942-ac6c-4ecb52f83a89

### Notes

`split.js` allows displaying the split in horizontal or vertical direction. However, it does not handle switching easily when changing between breakpoints. It does provide a `destroy` method which allows us remove _some_ event handlers and the attached `css` styles. The library does allow us to set the "sizes", but not direction.

So when we switch between horizontal & vertical display, we destroy the switch instance & re-initialise it with the new direction.

---

closes #1828


nokome:
Thanks @simonwinter looks great. I appreciate the extensive doc comments. I've merged in `main` to fix an issue and added a couple of `override` annotations.

It'll be nice to get some design refinements in here later (e.g. the split handle looks a bit like a scroll bar right now). Also the work that I am doing correlating the cursor position in the code editor to node ids will be be useful for linking the scroll positions in the two panels in the future.

nokome (approved):

Issue #1809, opened 2023-12-10
feat(views/source): syntax highlighting for Markdown extensions


nokome:
Stencila has several syntax extensions for Markdown. These need to be highlighted in the source code editor for documents. Each of the following extensions has Markdown examples in [`examples/nodes`](https://github.com/stencila/stencila/tree/main/examples/nodes).

- [ ] `Admonition`
- [ ] `IncludeBlock` and `CallBlock`
- [ ]  `Parameter`
- [x] `ForBlock` and `IfBlock`
- [ ] `StyledBlock` and `StyledInline`
- [ ] `InstructBlock` and `InstructInline`
- [x] `InsertBlock` and `InsertInline`
- [x] `ModifyBlock` and `ModifyInline`
- [x] `ReplaceBlock` and `ReplaceInline`
- [x] `DeleteBlock` and `DeleteInline`

PR #1846, opened 2023-12-17
Feature/stencila md syntax highlights 1809

mike-parkin:
** details **

create a syntax highlight system for the colon delimiter syntax `:::` this covers if / for and opne style blocks,

This does not create 'blocks' that wrap the entire content, as this will be very difficult to handle, specifically creating nested blocks and having to reparse the base markdown plus any other extensions withing these blocks, however a node will be created for each line with the delimiters, and should therefore be accessible in the Tree of the editors document.

have added some basic styling to the elements, however these are mearly placeholder stylings.

to run: will need to run `npm install` for a couple of new deps in the web dir

related issue: #1809


mike-parkin:
please feel free to change any names to something more appropriate, I used colonSyntax for the extension, but that's hardly the nicest name

Issue #1845, opened 2023-12-17
feat(python): add shortcut functions to create document content


nokome:
One use case for the Python SDK is to programmatically generate Stencila documents using Python (e.g. and then rendering them in the browser or exporting them to some other format). Currently the API for doing that is verbose and awkward e.g. 

```python
from stencila.types import Article, Paragraph, Text
article = Article(content=[Paragraph(content=[Text(value="Hello world!")])])
```

In Rust, we have create several [shortcut functions](https://github.com/stencila/stencila/blob/main/rust/schema/src/shortcuts.rs) which improve the developer experience for creating content nodes. Creating similar functions in Python would allow for a much more concise API e.g.

```python
from stencila.shortcuts import art, p
article = art(p("Hello world"))
```

This will require some logic to coerce arguments to the correct type. For example, the `p` shortcut function which creates a `Paragraph` which has `content` that is a list of `Inline` nodes, should be able to take as its arguments:

- a `list` of inline content e.g. `p(["Some ", em("emphasized"), " text"])`
- `*args` each of which is inline content e.g. `p("Some ", em("emphasized"), " text")`
- a `str` which is coerced to `[Text(value=arg)]` e.g. `p("Just a simple para")`

nokome:
Closed by #1851.

PR #1851, opened 2023-12-19
Python shortcuts

brettc:
WIP: Add shortcuts for python entity construction (#1845).

This also refactors the python schema generation to avoid name conflicts:
* Module names are now prefixed with an underscore (avoiding `list` and `object` coming into scope)
* We generate an `__all__` in `init.py` to allow for `from stencila.types import *`, which eases problems with forward references
* We allow for renaming of schema types to avoid name collisions (`List -> List_`)


brettc:
@nokome : Some updates on the PR

- I have copied all the shortcuts that I found in the `shortcuts.rs`. I *think* I got them right; but it was a long mind-numbing task!
- The shortcuts have been wrapped using beartype, which will give you some run-time type checking. i.e. They are safer to use than the originals, and will give sensible error reports if you don't get things right.
- I've tried to make things pythonic when possible, so you should be able to pass iterators / generators / lists of blocks as well as plain strings.
- There are some limited tests, which could use more work, especially since..
- ... I think stencila is returning some odd things for some Nodes (TableCell, for example).
- There are a few naming issues, where variable names and type names collide with python reserved names or builtins. I think I have cured the worst one (List/list), but there are others ('format', for example).
- I suspect the conversion code in`convert.py` is out of date (for schema updates). I think there is a way of simplifying this and avoiding the problem in the future...
- I've added the ruff linter configuration, but not finished incorporating it. This will highlight things like the above.

In any case, I think it can be merged, but I should problem add some more issues (will do, after you comment).

PR #1903, opened 2024-01-11
Expose more document state to `CodeMirrorClient`

nokome:
## The problem

The `CodeMirrorClient` provides a way of syncing between an in-browser representation of the document in some format (e.g. Markdown, JATS) with the document state on the server. However, formats such as Markdown are lossy in that they do not encode all the details of the state (e.g. `authors` and `executionStatus` properties). We want to be able to expose some of these details in the CodeMirror editor.

## The solution

- [x] Sync a JavaScript object representing the whole document between the server and the client. This object can be "reached into" to get the value of any property of any node in the document
- [x] Sync a mapping between character position and node id between the server and the client. This can be used to translate the current CodeMirror cursor position to a node in the document so we can do the "reaching in" to the JavaScript object
- [x] Add a method somewhere (on `SourceView` probably) which returns the document node for a given cursor position (i.e. combines the above two items)


nokome:
This is mostly complete now. In the browser, the most useful method is `SourceView.getNodesAt`.

nokome:
Remaining items to do before merging this..

- [ ] Check that mapping positions are for characters, not for bytes. Check that CodeMirror positions/ranges are for characters not bytes.
- Finish re-implementing `to_markdown`for:
   - [ ] Figure
   - [ ] InstructionBlock
   - [ ] InstructionInline
   - [ ] List
   - [ ] Note
   - [ ] Parameter
   - [ ] QuoteBlock
   - [ ] StyledBlock
   - [ ] StyledInline
   - [ ] Table

mike-parkin:
As of this morning the Table node and all its children are showing up in the object client, but only the Table nodes are being returned in the map, so when running getNodeAt() on a TableCell / TableRow we get the following image (this in for the Readme Btw)

image

image

PR #1905, opened 2024-01-14
refactor(instructSyntax): change syntax for instruct

mike-parkin:
** details ** 

Have updated the syntax highlighting for the instruction blocks from using the @@ to the %% %> %% style syntax for inertion and content editing


nokome:
Thanks for this Mike. I think we should hold off merging this until we consider changing the syntax again to be consistent with "slash commands" e.g `/ some instruction` and `{/ some instruction /> /}` 🤷🏽 .

PR #1919, opened 2024-01-19
Re-implementation of kernels and microkernels

nokome:
This brings across kernels and microkernels from the `v1` branch.

After reviewing `v1` I decided not to checkout -and-refactor that code but instead take the approach mostly taken elsewhere in `v2` of rewriting-with-liberal-copying. There is too much non-idiomatic Rust and other cruft to make the checkout-and-refactor approach worthwhile.


nokome:
@ignatiusm So far I've just done some initial scaffolding using the same crate organization and trait based approach we are taking for `codecs` and `assistants`. There is a mini-CLI as discussed in `rust/kernels/main.rs` but at this stage it does not do anything. 

I think it would be worthwhile waiting for me to do a little more scaffolding / setup before diving into this yourself. I might be able to get this later today, but more likely tomorrow.

We should setup a Markdown doc in `docs/reference/microkernels.md` and document as we go!

nokome:
@ignatiusm I've pushed up my latest work on this. I think it is ready for you to have a go at re-implementing a Python and/or R microkernel. Some things to note:

- there have been some changes to the protocol (where flags are located start/end of lines etc and possibly the unicodes used for each)
- I suggest starting off with a very simple re-implementation following the `kernel-bash` crate. I would `cp -r kernel-bash kernel-python` and then go around doing the appropriate renaming etc. That way you can reuse the tests and some of the other boilerplate.
- it might be a good idea for us to pair code on even this initial step (there are a bunch of gotchas that I can probably step you through to avoid frustration)
- at this stage I'm haven't reimplemented the forking stuff but I'll probably try to quite soon; so you can ignore that for now

PR #1959, opened 2024-01-25
Fix #1916 modify blocks

### LINKED ISSUE #1916: bug(assistants): `modify-blocks` or `modify-inlines` assistants types should create a `ReplaceBlock` or `ReplaceInline` suggestion
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction URL (if applicable)

_No response_

### branch

_No response_

### commit hash

_No response_

### Description

These two methods of `GenerateOutput` only return `InsertInline` or `InsertBlock`:

```rust
    /// Create a `SuggestionInlineType` from the output that can be used for the `suggestion`
    /// property of the instruction
    pub fn to_suggestion_inline(self) -> SuggestionInlineType {
        SuggestionInlineType::InsertInline(InsertInline {
            content: self.nodes.into_inlines(),
            ..Default::default()
        })
    }

    /// Create a `SuggestionBlockType` from the output that can be used for the `suggestion`
    /// property of the instruction
    pub fn to_suggestion_block(self) -> SuggestionBlockType {
        SuggestionBlockType::InsertBlock(InsertBlock {
            content: self.nodes.into_blocks(),
            ..Default::default()
        })
    }
 ```
 
They could instead be coalesced into one method with two args and insert `ReplaceInline` and `ReplaceBlock` when appropriate:

```rust
    /// Create a `SuggestionInlineType` from the output that can be used for the `suggestion`
    /// property of the instruction
    pub fn to_suggestion(self, inlines: bool, insert: bool) -> SuggestionInlineType {
        match (inlines, insert) {
            (true, true) => SuggestionInlineType::InsertInline(InsertInline {
                 content: self.nodes.into_inlines(),
                 ..Default::default()
              }),
             // etc
        }
    }
 ```

Then, when they are called, set these arguments based on the instruction type and whether it has any content e.g.

```rust
instruction.options.suggestion = Some(output.to_suggestion(true, instruction.content.is_some()));
```
 

### Reproduction steps

@brettc gave this example of something here is working on (note the `++` indicating a `InsertBlock` suggestion; should be `~~` for `ReplaceBlock`):

```
%% @local/shakespearize expand into two paragraphs.
%>

The south island of new zealand.

%%

++

The fairest isle of New Zealand in the south, Where rugged mountains kiss the sky so high, And verdant forests spread their leafy boughs, A land of wondrous beauty, none can deny.

In fields of green, the sheep doth gently graze, While crystal waters flow with soothing sound, This is a place of peace, a tranquil maze, Where nature's bounty doth abundantly abound.

Oh, New Zealand's south island, fair and grand, A paradise of nature's graceful hand.

++
```

### Browsers

_No response_

### Anything else?

_No response_

brettc:
This partially fixes #1916.
We still need to merge the instructions and the original text for a markdown.

### DIFF
--- a/rust/assistant/src/lib.rs
+++ b/rust/assistant/src/lib.rs
@@ -792,8 +792,14 @@ impl GenerateOutput {
 
     /// Create a `SuggestionInlineType` from the output that can be used for the `suggestion`
     /// property of the instruction
-    pub fn to_suggestion_inline(self) -> SuggestionInlineType {
-        SuggestionInlineType::InsertInline(InsertInline {
+    pub fn to_suggestion_inline(self, insert: bool) -> SuggestionInlineType {
+        if insert {
+            return SuggestionInlineType::InsertInline(InsertInline {
+                content: self.nodes.into_inlines(),
+                ..Default::default()
+            });
+        }
+        SuggestionInlineType::ReplaceInline(ReplaceInline {
             content: self.nodes.into_inlines(),
             ..Default::default()
         })
@@ -801,11 +807,17 @@ impl GenerateOutput {
 
     /// Create a `SuggestionBlockType` from the output that can be used for the `suggestion`
     /// property of the instruction
-    pub fn to_suggestion_block(self) -> SuggestionBlockType {
-        SuggestionBlockType::InsertBlock(InsertBlock {
+    pub fn to_suggestion_block(self, insert: bool) -> SuggestionBlockType {
+        if insert {
+            return SuggestionBlockType::InsertBlock(InsertBlock {
+                content: self.nodes.into_blocks(),
+                ..Default::default()
+            });
+        }
+        return SuggestionBlockType::ReplaceBlock(ReplaceBlock {
             content: self.nodes.into_blocks(),
             ..Default::default()
-        })
+        });
     }
 
     /// Display the generated output as Markdown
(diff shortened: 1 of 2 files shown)


brettc:
New issues to finish rendering markdown correctly #1960

rust/assistant/src/lib.rs
```diff
@@ -792,20 +793,32 @@ impl GenerateOutput {
...
+        if insert {
+            return SuggestionInlineType::InsertInline(InsertInline {
+                content: self.nodes.into_inlines(),
+                ..Default::default()
+            });
+        }
+        SuggestionInlineType::ReplaceInline(ReplaceInline {
+            replacement: self.nodes.into_inlines(),
             ..Default::default()
         })
```
nokome, line 806: Avoid early returns and instead just "return" the if expression:

```suggestion
        if insert {
            SuggestionInlineType::InsertInline(InsertInline {
                content: self.nodes.into_inlines(),
                ..Default::default()
            })
        } else {
            SuggestionInlineType::ReplaceInline(ReplaceInline {
                replacement: self.nodes.into_inlines(),
                ..Default::default()
            })
       }
```

rust/assistant/src/lib.rs
```diff
@@ -792,20 +793,32 @@ impl GenerateOutput {
...
+            return SuggestionBlockType::InsertBlock(InsertBlock {
+                content: self.nodes.into_blocks(),
+                ..Default::default()
+            });
+        }
+        return SuggestionBlockType::ReplaceBlock(ReplaceBlock {
+            replacement: self.nodes.into_blocks(),
             ..Default::default()
-        })
+        });
```
nokome, line 821: Same here, you can get rid of both these returns.

nokome (changes requested):
Thanks for this. Could you please address the minor style comments.

PR #1974, opened 2024-01-29
feat(bottom-panel): style bottom panel controls

### LINKED ISSUE #1958: feat(source view): bottom panel
### Summary

The `SourceView` needs some design! At present there are three things that we've just thrown in there in random places:

- dropdown for the format
- toggle for line wrapping
- breadcrumbs for current node

image

To keep things as visually uncluttered as possible, I suggest that as much as possible we put things into the bottom panel ala VSCode.

@kusalekanayake Could you do a design for that bottom panel please:

- Most of it will be straightforward I think and follow the lead of other editors.
- For the breadcrumbs I wonder about using the same icons for each type that we use elsewhere?
- It should be responsive e.g. hide breadcrumb labels on small screens

### Reference Issues

_No response_

### Implementation PR

_No response_

mike-parkin:
**details**

apply some styling to the bottom panel of the source editor, including toggle and active state for the word wrap button.

design:

image

implementation:

image

related issues: closes #1958

### DIFF
--- a/web/src/views/source/panels.ts
+++ b/web/src/views/source/panels.ts
@@ -64,8 +64,10 @@ class EditorPanelElement extends TWLitElement {
 
     const styles = apply(['w-28 h-full', 'mx-3 pl-2', 'bg-white', 'rounded-sm'])
 
+    const title = 'Select document format'
+
     return html`
-      <select title="Document format" class=${styles} @change=${changeEvent}>
+      <select title=${title} class=${styles} @change=${changeEvent}>
         ${Object.entries(FORMATS).map(
           ([format, name]) =>
             html`<option
@@ -83,11 +85,10 @@ class EditorPanelElement extends TWLitElement {
     const clickEvent = () => {
       this.sourceView.lineWrap = !this.sourceView.lineWrap
     }
-    const title = `Toggle line wrapping: ${this.sourceView.lineWrap ? 'ON' : 'OFF'}`
+    const title = `Turn ${this.sourceView.lineWrap ? 'off' : 'on'} line wrapping`
     const styles = apply([
       'h-4 w-4',
-      this.sourceView.lineWrap ? 'bg-gray-200' : '',
-      'hover:bg-green-000',
+      this.sourceView.lineWrap ? 'bg-gray-200' : 'hover:bg-green-000',
     ])
 
     return html`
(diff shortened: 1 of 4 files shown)


web/src/views/source/panels.ts
```diff
@@ -30,71 +31,81 @@ class EditorPanelElement extends TWLitElement {
...
+        <img src=${icon} width="100%" height="100%" />
+      </button>
     `
   }
 
   private renderBreadcrumbs() {
     return html`
-      <div>
+      <div class="text-xs leading-none flex items-center">
         ${this.breadcrumbs.reverse().map((entry, i, arr) => {
```
nokome, line 103:
```suggestion
        ${this.breadcrumbs.reverse().slice(1).map((entry, i, arr) => {
```

Don't include the first `Article` in the breadcrumb its redundant.

nokome (commented):
Thanks @mike-parkin. A couple of things:

- As per my comment, don't include the first `Article` in the breadcrumb, its redundant.
- Could you add padding to the left side of the language selector as in Kusal's design

nokome:
A bit more feedback after spinning this up locally:

- I think there needs to be more distinction between the toolbar and the source code (see below): more top padding on the bar?, a light grey top border on the bar?
image

- I think we should use Shoelace tooltips: the defaults esp on Firefox are ugly. I'd have them popping up above the control. 

[Screencast from 2024-01-30 08-50-14.webm](https://github.com/stencila/stencila/assets/1152336/df2abb00-4e60-49eb-82c0-1c6fae4555e6)

- The wording for the word wrap tool tip could be improved: "Turn on line wrapping" or "Turn off line wrapping". The tooltip for the format could also be reworded for consistency "Select document format".

- The shading of the word wrapping icon is confusing. It doesn't come on when you toggle on wrapping until you mouse off the icon. @kusalekanayake had a proposal for this sort of interaction in https://github.com/stencila/stencila/issues/1958#issuecomment-1913864101. Maybe see if he still wants to go with that.

mike-parkin:
Have pushed up the padding / Article / word wrap hover state 

Will change the wording and put a top border back in

 best to wait for some art direction on tooltips before implementing that component

kusalekanayake:
Planning on setting a consistent tooltip aesthetic soon. It's in the queue!

PR #1982, opened 2024-01-30
feat(design chrome): setting up app to reflect new chrome design

### LINKED ISSUE #1962: feat(design): chrome
### Summary

Update the chrome to reflect the new design (see below):

- [ ] Tool bar:
  - [ ] Views dropdown
  - [ ] Format dropdown
  - [ ] Sync status icon
  - [ ] info button
  - [ ] print button
- [ ] Side bar:
  - [ ] open / close file browser panel
  - [ ] settings
- [ ] File tabs
- [ ] Editor area

Source view insert menu

### Reference Issues

_No response_

### Implementation PR

_No response_

simonwinter:
This PR introduces the new chrome to the app!

We have:
- a toolbar
- a sidebar
- a content area

CleanShot 2024-01-30 at 17 49 50

## New things!

1. This introduces using shoelace. Currently only a few components are enabled, but can easily be added to in the `web/src/apps/shoelace.ts` file. Additionally, we have our own custom icon set that can easily be extended simply by dropping `svg` files into the `web/src/assets/icons` directory.
2. `stencila-ui-view-container` allows us to add in the content we want to view + a slot to add in components like the info box. If no additional components are needed the slot is hidden.
3. Styling based on views: the different views require slightly different styling, so we can now change things like height, scrolling etc by adjusting the appropriate classes of the view container.

## To do

There are still things I'm going to improve on:

- The theme css needs looking at, particularly for non-source code views.
- Connecting the buttons & managing what is/isn't active
- Add in the print preview modal
- Look at some better transitions & animations for effects like button hover states
- Connect the info box

The buttons currently don't do much as yet. They should enable us to connect the info panel, print view modal, sidebar (for selecting files) easily enough.

---

closes #1962


web/src/apps/shoelace.ts
```diff
@@ -12,4 +13,9 @@ import { version } from '../../package.json'
 const { NODE_ENV } = process.env
 const base = NODE_ENV === 'development' ? 'dev' : version
 
-setBasePath(`~static/${base}/shoelace-style/`)
+setBasePath(`/~static/${base}/shoelace-style/`)
```
nokome, line 16: OK, this maybe explains why we have to put the Shoelace CSS files in the `<head>`?

nokome (approved):
Thanks @simonwinter. Great to see this. A few minor comments from me.

mike-parkin:
this is more an issue to consider in #1958 and #1974 with the bottom panels

but will need to find a way extend out any codemirror elements that get chopped of by the codemirror boundaries

might just add a pseudo element to the bottom panel to extend out as far as needed :/

should probably check the scrollbar as well.

image

web/src/apps/shoelace.ts
simonwinter: I agree, it would be good to have rust ignore this. 

That's ☝️ partially the issue here. I haven't found a solution here I like - importing multiple css files into main causes some naming clash in parcel which other users had issues with as well. I would like to look at bundling `main.css` + other stylesheets into one ... I'd like to make a separate issue for this - it's not a show stopper right now, but should be addressed.

PR #1983, opened 2024-01-30
DRAFT: feat(assistants): use embeddings

brettc:
This shifts the specialised assistants to using examples with embeddings rather than regexes for matching.

- [x] Clean up assistant markdown using instructions and "expected-nodes" to simplify definition.
- [ ] Write tests to see how well the matching works (in progress)
- [ ] Write a matrix of embedding comparisons to ensure our examples are distinct enough (in progress, but requires refactor below)
- [ ] Refactor embeddings / cosine similarity (to discuss)
- [ ] Refine text based on tests (maybe leave to later)
- [ ] Incorporate (or move) the tests to use the examples/assistants (to discuss)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -252,6 +252,9 @@ dependencies = [
  "assistant-openai",
  "assistant-specialized",
  "codecs",
+ "serde",
+ "serde_yaml",
+ "tokio",
 ]
 
 [[package]]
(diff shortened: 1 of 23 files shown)


brettc:
Hey @nokome, please take a look and comment. Some extra info:

- The tests are currently in the "assistants/tests" folder. They are printing rather than asserting as this gives a better overview of what is going on for now (I can't see how to parametrise fixtures in rust!). i.e. `cargo test -- -nocapture`
- I've made a few things "pub" to get the tests working (it seems that non-unit tests don't have private access).
- The matrix is there, but getting the cosine similarity requires a task. I think it should be factored out into its own newtype which saves having Vec<Vec<f64>> everywhere, and the lazy init and cosine similarity can go into the impl.
- It definitely works, but a focused session on fine tuning is required.

PR #1996, opened 2024-02-01
fix(directory view): make it pretty

nokome:
Fixes up the `DirectoryView` to use the `ObjectClient` because using the `DomClient` didn't work with `<sl-tree>`.  In any case this is probably a better approach for this view.


nokome:
@simonwinter I've done some cleanup and added methods which can be used to implement actions such as opening a file, creating a directory or file etc. Also this can now be used without specifying the `doc` attribute. I just tried dropping into `<stencila-main-app>` like this and it would fine (although needs a loading spinner):

```html
<stencila-directory-view></stencila-directory-view>
```

PR #2020, opened 2024-02-11
Added more block icons and changed colours (small)

kusalekanayake:
Minimal changes to add a couple of icons for some of the blocks and altered with the colours.

Want to see how we can use the tailwind colours however since the default green is of a different tone to the one Stencila should be using.


web/src/nodes/helpers/node-ui.ts
```diff
@@ -35,21 +35,21 @@ const nodeColours = (name: string) => ({
...
+  Heading:          { ...stencilaIcon('heading'),              ...nodeColours('purple')  },
   IfBlock:          { ...stencilaIcon('if-block'),             ...nodeColours('amber')   },
   InsertBlock:      { ...shoelaceIcon('plus-circle'),          ...nodeColours('lime')    },
-  InstructionBlock: { ...stencilaIcon('instruction-block'),    ...nodeColours('yellow')  },
-  List:             { ...stencilaIcon('list'),                 ...nodeColours('orange')  },
-  MathBlock:        { ...shoelaceIcon('infinity'),             ...nodeColours('fuchsia') },
+  InstructionBlock: { ...stencilaIcon('instruction-block'),    ...nodeColours('blue')  },
+  List:             { ...stencilaIcon('list'),                 ...nodeColours('yellow')  },
+  MathBlock:        { ...stencilaIcon('math-block'),           ...nodeColours('fuchsia') },
   Paragraph:        { ...stencilaIcon('paragraph'),            ...nodeColours('green')   },
```
nokome, line 50: Somewhere else (in Discord maybe) I proposed using more neutral colors such as grey or a blue for paragraphs since there will be a lot of them and they are less "special" than things like instructions etc

nokome (changes requested):
Thanks @kusalekanayake. Could you see what you think of my comments. Happy to discuss.

web/src/assets/icons/replace-block.svg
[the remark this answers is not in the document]
kusalekanayake: Sounds appropriate!

web/src/nodes/helpers/node-ui.ts
kusalekanayake: Yep just caught that comment. ALso for the toggle, that's where I thought the 
image
icon would function for. Since it is showing information about the document blocks. Potential for the future to also have a section dedicated to overall document info.

web/src/assets/icons/math-block.svg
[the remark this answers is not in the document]
kusalekanayake: Fair! That make sense.

web/src/nodes/helpers/node-ui.ts
kusalekanayake: Do we have an idea on what blocks specifically we could consider more "common" or "normal"? Just trying to figure out where to draw the line.

nokome: > The initial thought was to give most of the basic blocks that feint green as a base a it's one of Stencilas main brand colours, so having it sprinkled more commonly throughout the app would give it a bit more of a connection to the brand. Making it the blue could also work, since we can associate green with things like blocks that are associated with addition.

I think that's good reasoning. I'd lean towards using the blue tones for the most basic blocks and the green for suggested additions.

> Do we have an idea on what blocks specifically we could consider more "common" or "normal"? Just trying to figure out where to draw the line.

I suspect a rough order or common-ess would be: paragraphs, headings, lists, tables, figures. All the rest feel more "special" to me so definitely the more vibrant colors there.

nokome:
Thanks for this @kusalekanayake , have merged now

PR #2026, opened 2024-02-13
feature: add kernel-r

ignatiusm:


rust/kernel-r/src/lib.rs
```diff
@@ -0,0 +1,835 @@
...
+/// A kernel for executing R code
+#[derive(Default)]
+pub struct RKernel {
+    /// A counter of instances of this microkernel
+    instances: AtomicU64,
+}
+
+impl Kernel for RKernel {
+    fn id(&self) -> String {
+        "R".to_string()
```
nokome, line 17: For consistency with the other kernels, I think this should be lowercase `r`

rust/kernel-r/src/lib.rs
```diff
@@ -0,0 +1,835 @@
...
+        self.microkernel_supports_terminate()
+    }
+
+    fn supports_kill(&self) -> KernelKill {
+        self.microkernel_supports_kill()
+    }
+
+    fn supports_forks(&self) -> KernelForks {
+        // Uses R `os.fork()` which is only available on POSIX-based systems
+        // TODO: fix for R
```
nokome, line 42: See the implementation of `microkernel_supports_forks` suggest this change:

```suggestion
        // not Windows so use `microkernel_supports_forks()`
```

nokome (commented):
I realize this is still in very early stages but thought I would add some comments.

rust/kernel-r/src/lib.rs
```diff
@@ -0,0 +1,835 @@
...
+    #[tokio::test]
+    async fn messages() -> Result<()> {
+        let Some(mut kernel) = start_instance::<RKernel>().await? else {
+            return Ok(());
+        };
+
+        // Syntax error
+        let (outputs, messages) = kernel.execute("bad ^ # syntax").await?;
+        assert_eq!(messages[0].error_type.as_deref(), Some("SyntaxError"));
+        assert_eq!(messages[0].error_message, "invalid syntax (<code>, line 1)");
```
nokome, line 228: There was a recent change in the schema for messages. This should be:

```
messages[0].message
```

rust/kernel-r/src/lib.rs
```diff
@@ -0,0 +1,835 @@
...
+        assert!(messages[0].stack_trace.is_some());
+        assert_eq!(outputs, vec![]);
+
+        // Runtime error
+        let (outputs, messages) = kernel.execute("foo").await?;
+        assert_eq!(messages[0].error_type.as_deref(), Some("NameError"));
+        assert_eq!(messages[0].error_message, "name 'foo' is not defined");
```
nokome, line 235: Ditto

nokome:
Merged although tests still need to be enabled on CI. See #2078

Issue #2039, opened 2024-02-18
chore(python kernel): review initial implementation


nokome:
There is a initial (re)implementation of a [Python kernel](https://github.com/stencila/stencila/blob/d5299bf021ca56a22dbbb60af477f2151deac0af/rust/kernel-python/src). This was developed in parallel with kernels for Bash, Node.js and Rhai (primarily to ensure the API for kernels was general enough to cover all). There is a kernel for R [in development](https://github.com/stencila/stencila/pull/2026).

It would be good to have a Python expert review the [kernel.py](https://github.com/stencila/stencila/blob/d5299bf021ca56a22dbbb60af477f2151deac0af/rust/kernel-python/src/kernel.py) making it more idiomatic where appropriate. There are few things to consider in this review:

- We have tried to strike a balance between well documented code and size of code: this file gets embedded into the `stencila` binary. So, while still a very small part of the total size of the binary, it is good not to introduce unnecessary bloat.

- This aims to run on reasonably old versions of Python3 so we have avoided using typing and other more modern, less well support, Python features

- One of the reasons for doing this review is so that @brettc can become familiar with how we doing kernel introspection to add to the context of assistants so they can generate more useful code

PR #2048, opened 2024-02-19
Python kernel review

brettc:
PR for #2039 

Clean-up tasks done. This is mostly just tidying, following some sensible rules:

- [x] Add ruff.toml with set of rules to check.
- [x] Add ruff checking to Makefile
- [x] Reformat code to Black standard (88 chars)
- [x] Use dict literals and dict comprehensions
- [x] Remove unnecessary elsif's
- [x] Move loop code to "main".

Possible things to do:

- ~~[ ] Replace chained elif with a dictionary/lambdas~~
- [x] Add minimal typing (to aid readability and ensure consistency)
- ~~[ ] Review the matplotlib status -- it assumes some conventions that I'm not sure are always the case.~~

### DIFF
--- a/rust/Makefile
+++ b/rust/Makefile
@@ -31,6 +31,8 @@ fix:
 # Run linting checks
 lint:
 	cargo clippy --workspace --tests
+	ruff -n kernel-python/src/kernel.py
+	pyright kernel-python/src/kernel.py
 
 # Setup before running tests
 test-setup:
--- a/rust/kernel-python/ruff.toml
+++ b/rust/kernel-python/ruff.toml
@@ -11,31 +11,29 @@ target-version = "py39"
 # Unlike Flake8, Ruff doesn't enable pycodestyle warnings (`W`) or
 # McCabe complexity (`C901`) by default.
 select = [
-  'F', # pyflakes
-  'E', # codestyle
-  'W', # warning
-  # 'C90', # complexity
-  'N',  # pep-naming
-  'UP', # upgrade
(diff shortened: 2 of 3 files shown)


nokome (commented):
Thanks for this @brettc. Looks good, some minor comments/queries:

> Add ruff checking to Makefile

I think this should go under `rust/Makefile`s `fix` recipe?

> Replace chained elif with a dictionary/lambdas

I'm curious what the advantage of this is, I thought it would be less performant?

>  Add minimal typing (to aid readability and ensure consistency)

+1

> Review the matplotlib status -- it assumes some conventions that I'm not sure are always the case.

Yes please!

brettc:
> 
> I think this should go under `rust/Makefile`s `fix` recipe?
> 

I put it under `lint` (it does not fix anything). I've added pyright too, which I like more than mypy.
They will require installation, but they are both able to be installed easily.

> > Replace chained elif with a dictionary/lambdas
> 
> I'm curious what the advantage of this is, I thought it would be less performant?
> 

I've left this for now. In general, I find long chains of elifs error-prone, and a dict/func avoids some problems. I thought it would be faster. 

> > Add minimal typing (to aid readability and ensure consistency)
> 

Done!

> > Review the matplotlib status -- it assumes some conventions that I'm not sure are always the case.
> 

I'm going to create another issue for this. It requires some investigation.

nokome:
Thanks! Have added `ruff` and `pyright` to CI so linting passes there & will merge.

Issue #2049, opened 2024-02-19
bug(writer document tabs): styling and vertical spacing


nokome:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

Shoelace tabs have been added but they have not been styled as per design and to avoid bottom toolbar of editor going off screen the header has been made too narrow:

![image](https://github.com/stencila/stencila/assets/1152336/d6fe2e87-dab9-483e-8dc9-ad3b64d6392b)

nokome:
Closing as this app skeleton no longer used (#2309).

PR #2062, opened 2024-02-20
Feature/style doc tabs 2049

mike-parkin:
**details**

Have applied new styles to the tabs to match the designs.

And the bottom panel is back to where it was before.

there are some bugs with the tab functionality which i believe should be a created as a separate issue.

**related issues**: #2049


nokome:
Thanks @mike-parkin. I span this up and took it for a test drive. Looks fine to me but think that it could do with some tweaks and we should review as part of a UI/UX audit:

image

PR #2067, opened 2024-02-21
Feature/source writeonly mode 2052

### LINKED ISSUE #2052: feat(source editor): write-only mode
## Problems

Currently, the editor in the `SourceView` is write-read: the user is able to modify the source (e.g. Markdown) in the editor, those updates get sent to the server as patches and merged into the in-memory document node tree; the source is then regenerated on the server based on the updated tree, and changes to the source (and to the mapping between character position and node) are sent back to the client. This bidirectional flow allows reactivity (e.g for AI suggestions to appear in Markdown) and for real-time collaboration on source formats (because each editor gets updates to the source by each other).

There are however several problems with the write-read bidirectional flow:

1. Automatically adding new content, such as AI suggestions, to the Markdown breaks the usual expectations about how a _source_ editor works

2. There are differences between the user written Markdown and the generated Markdown (e.g. multiple blank lines between paragraphs) which causes jumpiness which currently makes the editor unusable.

3. The fact that JavaScript uses UTF-16 and Rust uses UTF-8 for strings is breaking the offsets sent in patches between the two languages

3 is fixable (~ 1 day of work), 2 is mostly fixable (e.g. by modifying the Markdown parser to consider whitespace significant by parsing to empty paragraphs, ~ 1 day work). 1 is user related: some users just won't like Markdown getting inserted into their source.

## Proposal

- Allow the user to switch between write-only and write-read modes in the source editor. In write-only mode, the node type and node status gutters and the node cards would not be available (because the we are no longer able to provide a mapping between positions in the source and nodes)

- Provide and indicator to the user that the source has been updated by another actor (e.g. by an assistant, by another remote user) so that they can explicitly reload the latest version of the source when they want

-  In split mode, make write-only the default: the user can preview the document, including content added by AI assistants, in the right panel and click on nodes to get a node card in the info panel.

### Reference Issues

_No response_

### Implementation PR

_No response_

mike-parkin:
**details**

Create a `writeOnly` setting for the source editor,

When active this disables the source views `codeMirrorClient` object from dispatching transactions to the editor,

It also disables the infoBox and statucGutter editor extensions, which have been wrapped up in a codemirror`Comparment` called `clientRecieverConfig` to enable smoother disabling without having to rerender the entire editor.

Have added a toggle switch for this mode in the bottom toolbar,
there is currently a placeholder icon for the button, which is quite ugly, but it will at least remind us to put in a real one.

image

**related issue**: closes #2052


web/src/clients/codemirror.ts
```diff
@@ -59,15 +59,31 @@ export class CodeMirrorClient extends FormatClient {
    */
   private currentSelection?: SelectionRange
 
+  /**
+   * If true the client will not send messages from the server
+   * back to the `editor`
+   */
+  private _writeOnly: boolean
```
nokome, line 66: We should be consistent with use of underscores for private variables. I'd prefer to not use them, but happy to be outvoted if we are consistent. Is there an eslint for that?

web/src/views/source/bottom-panel.ts
```diff
@@ -55,11 +55,12 @@ class PanelElement extends LitElement {
     return html`
       <div class=${styles}>
         ${this.renderLineWrapButton()} ${this.renderFormatSelect()}
+        ${this.renderWriteOnlyButton()}
```
nokome, line 58: I reckon this new button should go after the existing one.

nokome (changes requested):
Thanks @mike-parkin. Looks good to me but a couple of minor comments.

web/src/clients/codemirror.ts
mike-parkin: fair enough, I think there's a leadingUnderscore setting in the naming conventions rule. But happy to remove that underscore for now.

nokome: Thanks

PR #2068, opened 2024-02-21
feat(infoview): view for information on document including currently selected node

### LINKED ISSUE #2046: bug(split view): extra vertical scroll bar, duplicated node card
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction URL (if applicable)

_No response_

### branch

_No response_

### commit hash

_No response_

### Description

In split view we get this:

image

I suggest that we need to do some refactoring here: remove the "info sidebar" from `SourceView` and make create a new `InfoView` (which will get toggled on/off by the info icon in top right, and slide out from the right). This view will be responsible for rendering summary info about the document as a whole (e.g. AI usage etc) and pop-up "node cards" for the currently selected node type, regardless of which view is in the centre - visual editor, source editor, or split view.

### Reproduction steps

_No response_

### Browsers

_No response_

### Anything else?

_No response_

nokome:
Closes #2046.

I have set up the basic scaffolding for this view. That includes removing similar functionality from the `SourceView` and creating an `InfoView`. I have put the custom element `<stencila-info-view>` into the tab panel for each document. 

I haven't done any CSS/design related stuff other than sketching out the approach to showing the selected node by changing its class to `show`. See comments as to why I think this is a better approach than we had before.

The `InfoView` will have to listen to events from the sibling `SourceView`, `VisualView` etc as to which node is currently selected. Or maybe a context object is needed here?

We should keep in mind that this view will be reused in the read app for implementing Kusals most recent designs. So it's not just about showing the currently selected node. It'll also show other summary information about the doc.

Handing over to you @simonwinter. Let me know if you have any questions.


simonwinter:
Is anyone able to help with these merge conflicts? @nokome @mike-parkin?

mike-parkin:
> Is anyone able to help with these merge conflicts? @nokome @mike-parkin?

not sure about the package.lock file
But otherwise, It looks like you can accept all the incoming changes

nokome:
This PR has been superseeded so closing

PR #2069, opened 2024-02-21
feat(plugins): establish plugin registry and API

nokome:
This PR aims to implement an initial iteration of a plugin registry and API. Still a very much a work in progress but aiming to include this early version in this week's alpha release of the CLI.

At this stage there is a `plugins.toml` file in the root of this repository which acts as a registry and points to URLs containing the `manifest.toml` file of each plugin. I have created two example plugins, each with their own repo:

image

Moving on next to plugin installation and fleshing out the manifest more:

- [x] Determine if a plugin is installable on machine (runtime available and platform supported) or if already installed
- [x] Do a fake install of a plugin by creating a dir for it and copying manifest there
- [x] Determine if the installed version of the plugin is out of date (compared version in 'manifests.json' with version in 'manifest.toml'

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -95,6 +95,7 @@
         "rfind",
         "rhai",
         "rowspan",
+        "runtimes",
         "rustfmt",
         "schemars",
         "serde",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3090,6 +3090,7 @@ dependencies = [
  "kernel-python",
  "kernel-r",
  "kernel-rhai",
+ "plugins",
 ]
 
 [[package]]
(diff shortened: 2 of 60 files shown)


nokome:
@brettc Could you jump on to this when you have time.

I suggest starting by trying out the existing CLI and installing/uninstalling the `node-example` plugin

```console
cargo run -p cli plugins --help
```

Then if you could implement a `class Plugin` in the https://github.com/stencila/plugin-example-python repo that is similar to the same class in https://github.com/stencila/plugin-example-node/blob/main/index.ts (I'd kick start that by asking your favourite LLM to convert that class to a Python class :)

Issue #2065, opened 2024-02-20
bug(writer tabs): glitches associated with tabs


nokome:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Description

There are glitches associated with opening and closing document tabs, including but perhaps not limited to:

- if a file e.g. `README.md` is opened automatically (because it is the default "index" file) and you click on it in the file explorer a duplicate tab is opened, instead of just making the existing tab active (if not already)

- when closing tabs sometimes the tabs are not associated with the panels properly, I suspect this bit of code may be to blame

```ts
    // Event listener for closing a document
    // If the closed document is the active document, makes the document
    // before it the active document (unless closing the very fast document)
    this.shadowRoot.addEventListener(
      'stencila-close-document',
      async (e: Event & { detail: { docId: DocumentId } }) => {
        const { docId } = e.detail

        const index = this.docs.findIndex((doc) => doc.docId === docId)
        if (index > -1) {
          this.docs.splice(index, 1)
          if (docId === this.activeDoc) {
            if (index == 0) {
              this.activeDoc =
                this.docs.length === 0 ? null : this.docs[0].docId
            } else {
              this.activeDoc = this.docs[index - 1].docId
            }
          }
          this.requestUpdate()
        }

        // Close the document on the server. This is a clean up task
        // which should not block the rendering so is not awaited here.
        DirectoryView.closeDocument(docId)
      }
    )
```

mike-parkin:
There is a PR up that should resolve this one:

 - when closing tabs sometimes the tabs are not associated with the panels properly, I suspect this bit of code may be to blame. 
 
https://github.com/stencila/stencila/pull/2071
 
 
The cause if this one here:

- if a file e.g. README.md is opened automatically (because it is the default "index" file) and you click on it in the file explorer a duplicate tab is opened, instead of just making the existing tab active (if not already)

When you initially load the page, the README.md is pulled in with an absolute filepath: `/home/users/mike/.../stencila/README.md`, and when you select from the directory view, if is a relative path, hence why it is not treating them as two different files.

nokome:
Thanks for diagnosing all this @mike-parkin! I've fixed the issue with the server starting up with absolute paths. However, there still seem to be various tabbing glitches. Will close this now and ask @kusalekanayake to do a UX audit to identify new issues.

PR #2071, opened 2024-02-22
fix(views): set views to respond to doc property update

mike-parkin:
**details**

This fixes an issue related to the document tabs in all views,

 - when closing tabs sometimes the tabs are not associated with the panels properly
 
The issue was not with the logic, the view in the panels were not updating itself
 
TL:DR; 
Here's what was happening: 
If i had two tabs open with ids: doc_1 and doc_2.
and I closed doc_1 and the main LitELement updates,  doc_2 is now the only tab, however it has moved into the panel where doc_1 was before. The property `doc` was updated correctly and passed down to child elements, however the view (eg, source, visual) element inside that panel had no way of responding an update in the `doc` property. I have set it now so It reset each view in this case.

**related issue**: #2065

### DIFF
--- a/web/src/views/dynamic.ts
+++ b/web/src/views/dynamic.ts
@@ -52,19 +52,21 @@ export class DynamicView extends ThemedView {
    * Override so that clients are instantiated _after_ this
    * element has a document `[data-root]` element in its `renderRoot`.
    */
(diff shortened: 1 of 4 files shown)


nokome (approved):
Thanks @mike-parkin. Looks good to me, will merge in now and work on the "README.md" issue separately.

Issue #2066, opened 2024-02-21
feat(rest api calls): consume fetch calls via a client


simonwinter:
### Summary

To help manage our fetch calls to internal API routes, we will have a `RestAPIClient` that will provide access to specific endpoints.

Initially, this is to support: #2060 & more specifically #1909 

- [x] create the `RestAPIClient` class
- [x] set up initial `listSecrets` function
- [x] define function Type
- [x] set up initial response payload (can include response or error)

@nokome This will be an initial stab & thoughts around particularly the payload are welcome!

nokome:
Here's my first blush thoughts on the API including copypasta from functions in `directory.ts` which could be replaces with this client

```ts
import { DocumentId } from '../types'

type Secret = {
  category: 'AiApiKey'
  name: string
  title: string
  description: string
  redacted: string
}

export class RestAPIClient {
  /**
   * Open a document by path on the server
   *
   * Returns the id of the document which can be used to open a view for the
   * document (by setting the `doc` attribute of the view element) e.g.
   *
   *   const view = document.createElement('stencila-live-view')
   *   view.setAttribute('doc', docId)
   */
  async openDocument(path: string): Promise<DocumentId> {
    const response = await fetch(`/~open/${path}`)
    if (response.status !== 200) {
      // TODO: Better error handling
      console.error(response)
    }
    const doc = await response.json()
    return doc.id
  }

  /**
   * Close a document on the server
   */
  async closeDocument(docId: DocumentId): Promise<void> {
    await fetch(`/~close/${docId}`, { method: 'POST' })
  }

  /**
   * List secrets
   */
  async listSecrets(): Promise<Secret[]> {
    const response = await fetch(`/~secrets`)
    // TODO: error handling
    return (await response.json()) as Secret[]
  }

  /**
   * Set a secret
   */
  async setSecret(name: string, value: string): Promise<void> {
    const response = await fetch(`/~secrets/${name}`, {
      method: 'POST',
      headers: { 'Content-Type': 'text/plain', body: value },
    })
    // TODO: check response code is 201
  }

  /**
   * Delete a secret
   */
  async deleteSecret(name: string): Promise<void> {
    const response = await fetch(`/~secrets/${name}`, { method: 'DELETE' })
    // TODO: check response code is 204
  }
}
```

PR #2061, opened 2024-02-20
feat(secrets): add support for secrets management

nokome:
Closes #1909

Still to do:

- [x] Add `Description` column to CLI table for `stencila secrets list`
- [x] Restrict setting and deleting of secrets to know secrets (in function and in CLI possible values)
~~- [ ] Add `SecretsClient` in `web` to use REST API exposed by server~~
- [ ] Implement UI as in #1909

PR #2072, opened 2024-02-23
fix(config): added rest api client class and config ui

### LINKED ISSUE #2060: feat(config): UI from Rust config struct
In #2027 I suggested that we use an "adaptive" approach to implementing a UI for changing config settings:

> For changing configuration settings of the app. Here for completeness but I don't think we should spend time on designs here. Rather, I think we should use something like [this](https://github.com/json-schema-form-element/jsfe) to take a JSON Schema specification of the config and automatically generate a form for it. The design input required here will be to audit the generated UI for usability and consistency with the rest of the app.

Elsewhere @simonwinter noted this:

> Front end look at: https://jsfe.js.org/Sink/AllFeatures 

 _Originally posted by @simonwinter in [#1909](https://github.com/stencila/stencila/issues/1909#issuecomment-1953371466)_

The first steps to implementing this are:

1. Define the config struct in Rust
2. Generate the JSON Schema from the struct using `schemars`

simonwinter:
**Description:**

This PR provides:

- [x] a `RestAPIClient` that provides a place to put all internal REST calls
- [x] Sets up a UI to display the found config

**TODO:**

- [x] To work with shoeslace's icons, we need to have the API icons as SVG
- [x] Need enable toggling of the config panel via the sidebar (update the sidebar's context)
- [x] Need to add ability to save changes to the API keys
- [x] Need to fix the discard button styles

**Future work:**

These are all likely issues for future development:

- [x] Need to break the UI up into smaller components - there's a lot going on in this 1 component!
- [x] We should track changes to the API fields, when the user saves each of these are saved to the server
- [ ] Need to implement @kusalekanayake's error handling designs #2093 
- [ ] Need to add animated logo to save button #2094 
- [ ] Need to implement clearable field designs #2095

---
closes #2060 #2061 #2066


web/src/assets/icons/other-api-LOGO.png
nokome, line 1: We're being a bit conservative, for security reasons, and only allowing specific API keys to be set as secrets through stencila. See https://github.com/stencila/stencila/pull/2061#issuecomment-1955599879 for the current list.

web/src/clients/RestAPIClient.ts
```diff
@@ -0,0 +1,226 @@
...
+   * Open the document at the supplied _path_
+   *
+   * @static
+   * @param {string} path
+   * @return {*}
+   * @memberof RestAPIClient
+   */
+  static async openDocument(path: string) {
+    return requestApiResponseAsJson<{ id: DocumentId }>({
+      url: `/~open/${path}`,
```
nokome, line 166: On the `feature/source-view-run` this has been moved to `/~documents/open/${path}`

web/src/clients/RestAPIClient.ts
```diff
@@ -0,0 +1,226 @@
...
+  }
+
+  /**
+   * close a document on the server
+   *
+   * @param {DocumentId} docId
+   * @memberof RestAPIClient
+   */
+  static async closeDocument(docId: DocumentId) {
+    await requestApi({ url: `/~close/${docId}`, init: { method: 'POST' } })
```
nokome, line 177: On the `feature/source-view-run` this has been moved to `/~documents/${docId}/close`

web/src/ui/config/icons.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Define the acceptable icon values that are available to the config screen.
+ */
+
+export type ICON_KEYS =
+  | 'ANTHROPIC_API_KEY'
+  | 'GOOGLE_AI_API_KEY'
+  | 'OPENAI_API_KEY'
+  | 'OLLAMA_API_KEY'
```
nokome, line 9: As per above comment, not needed.

```suggestion
```

web/src/ui/config/icons.ts
```diff
@@ -0,0 +1,18 @@
...
+  | 'MISTRAL_API_KEY'
+
+export const API_ICONS = {
+  ANTHROPIC_API_KEY: 'settings',
+  GOOGLE_AI_API_KEY: 'google-LOGO',
+  OPENAI_API_KEY: 'open-ai-LOGO',
+  OLLAMA_API_KEY: 'ollama-LOGO',
```
nokome, line 16: Ditto

nokome (commented):
I realize this is still DRAFT but I've left some comments in case they save some time. 

Re SVG v PNGs for icons. Can't we just use the PNGs in ``s, what advantage do the Shoelace icons give us in this situation?

nokome (approved):
Looks great, thanks @simonwinter.

Issue #2043, opened 2024-02-19
bug(source view): node type gutter indicator glitches


nokome:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Reproduction URL (if applicable)

http://localhost:9000/nodes/article-ark/article-ark.json?view=source

### branch

main

### commit hash

804069e0abdf1790c6534009674b62e0c589a036

### Description

Some examples from `examples/nodes/article-ark/article-ark.json`:

### Duplicated icons for `CodeBlock`, `CodeChunk`, `MathBlock`, (others?)

![image](https://github.com/stencila/stencila/assets/1152336/6f2c1aee-6dc1-4148-a3ad-90bb606b00c0)

### When line wrapping is on, gutters works,... until it doesn't

I'm not sure it this is easily fixable. It may not be worth fixing if it isn't as it only seems to appear at really narrow widths:

[Screencast from 2024-02-19 15-53-00.webm](https://github.com/stencila/stencila/assets/1152336/58109df3-e5e2-4474-a1de-5a45b248811b)

### Broken range for `QuoteBlock`

I'm putting this here for completeness but suspect it is an issue with the calculation of character ranges for `QuoteBlock` in the Rust encoding to Markdown:

![image](https://github.com/stencila/stencila/assets/1152336/31a5a0f5-5ab3-42e4-b326-7cfc89ba2a0c)

mike-parkin:
If we see these logs here relelated to the code block,

We can see that the lines with opening and closing delimiters are using the same start / end values

however the line in between with the 'code' content of the block, is also a `CodeBlock` node containing the `property: "code"` field and start\end values only cover the position of the content on that line. 

This is the same for the `MathBock` and the `CodeChunk` in the same example doc (article-ark.md).

![image](https://github.com/stencila/stencila/assets/73506758/9f58cb9b-08e7-429f-9d0e-325bbc8722c5)

![image](https://github.com/stencila/stencila/assets/73506758/8d9b5495-b143-4dd5-9e1f-0b529d252a4b)

nokome:
> however the line in between with the 'code' content of the block, is also a `CodeBlock` node containing the `property: "code"` field and start\end values only cover the position of the content on that line.

That is the intended behaviour: each `MappingEntry` provides the start and end character for either the node as a whole, or if `property` is defined, for the property of that node. So for this use case, the gutter should only use entries where there is no property.

mike-parkin:
TL:DR;
In relation to the above: 
[...]

Issue #2042, opened 2024-02-19
bug(source view): node type gutter does not show until editor is activated; need toggle


nokome:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### branch

main

### commit hash

804069e0abdf1790c6534009674b62e0c589a036

### Description

Current the node type gutter in the `SourceView` does not appear until the editor is activated (but sometimes it does).  Also there should be a toggle, with tool tip, in the bottom panel to show/hide the gutter.

nokome:
No longer relevant, removed.

PR #2084, opened 2024-02-27
Feature/toggle gutter markers 2042

### LINKED ISSUE #2082: (design)Icons for left gutters and write mode
**Summary**
We need icons for the left gutter where the icons and spans for the blocks are shown.

They need to be toggle-able and located in the right area.

Image

mike-parkin:
**details**

Added toggle switch for the gutter markers extension to the source view,

- the switch is in the bottom toolbar.
- It will be disabled when the 'writeonly' mode is on
- added custom icon for this (and the writeonly mode as well)

Have fixed the glitch of the gutters, with the code/math style blocks,

Still to be resolved: major gutter glitches with QuoteBlocks, still some minor glitches with ListBlock, Some files not loading gutters in editor until user interacts with editor.

**Related Issues**:  #2042 #2043  closes #2082


web/src/views/source.ts
```diff
@@ -91,6 +91,13 @@ export class SourceView extends TWLitElement {
   @property({ attribute: 'write-only', type: Boolean })
   writeOnly: boolean = false
 
+  /**
+   *  Turn on/off the node gutter markers.
+   *  Gutters will be disabled automatically in the "wite-only" mode.
```
nokome, line 96:
```suggestion
   *  Gutters will be disabled automatically in "write-only" mode.
```

nokome (approved):

web/src/views/source/bottomPanel.ts
```diff
@@ -113,6 +110,28 @@ class PanelElement extends LitElement {
     `
   }
 
+  private renderFormatSelect = () => {
+    const changeEvent = (e: Event) =>
+      (this.sourceView.format = (e.target as HTMLSelectElement).value)
```
simonwinter, line 115: Entirely an opinion: does anyone like the parenthesis style function body? For readability I believe that block style functions offer more clarity.

simonwinter (changes requested):
Minor comment to review. Nothing wrong as such - just an opinion offered & perhaps if we feel it necessary, add a rule to linting.

web/src/views/source/bottomPanel.ts
nokome: I'm not a fan of the readability of arrow functions generally but agree the block style is slightly less worse than parentheses.

PR #2087, opened 2024-02-29
Rework python schema

brettc:
- [x] fix bugs in shortcuts
- [x] deal with List/list problems
- [x] add typing for stencila / rust modules
- [x] switch to google doc format in convert
- [x] simplify from_value
- [x] test from_value still works
- [x] add some more tests


nokome:
On `main` I've had to skip tests because the `Article.__init__` method is broken:

image

It is broken because in the schema, I also added `Executable` to its base classes and the `schema-gen/src/python.rs` module assumes everything only extends one other type. So the generation function needs to handle the fact that `Article` now has two bases.

image

brettc:
It might pay to use this branch to upgrade the whole schema-gen. I've looked into how we might do something that will work for python 3.10 onward. This includes some upgrade to dataclasses, and better support for forward annotations (though the whole thing is broken and looks like it is finally getting fixed properly in python 3.13!).

According to some experimentation, the following should work

* Continue to use dataclasses, but use kw_only = True.
* Multiple inheritance appears to work okay.
* use `from __future__ import annotations`
* use native dataclass init (rather than overriding it as is done currently).
* IMPORTANT: define all the classes in a single file. This removes some of pythons deeper problems with forward annotations (TLDR; get_type_hints(), required for good type support, breaks across modules in many cases, and this solves it)
* use `cattrs` to dump and load json classes (rather than the attrs-json mixin). This also supports other serialisation (cbor etc, *without* needing stencila/rust support.

nokome:
I've fixed up some of the remaining issues around dependencies and merged.

PR #2096, opened 2024-03-04
feat(assistants): use document context to improve suggestions

nokome:
This PR add a document `Context` which is populated as the `Executor` walks over the node tree and can be used in the system prompt template of specialized assistants to improve their accuracy and reduce hallucinations.

This PR also removes some of the commands that we had in the CLI for testing assistants and instead improves the `execute` command so that a document can be executed and outputed to another file or to terminal in a format of choice.

I've added a example document `assistants/insert-code-chunk/variables-python.md` as an illustration of the types of examples we should add and test. This one in named as such because it tests the ability to get information about variables (in this case the columns in a pandas dataframe read from a CSV file) to an assistant. Given the brevity of the instruction in this examples, and that this is very much a first stab, the results are encouraging:

```sh
cargo run -p cli execute examples
/assistants/insert-code-chunk/variables-python.md --to md
```

````markdown
---
executionDuration:
  type: Duration
  value: 17194
  timeUnit: Millisecond
executionEnded:
  type: Timestamp
  value: 1709581296101
  timeUnit: Millisecond
executionCount: 1
executionRequired: No
executionStatus: Succeeded
---

```python exec
import pandas as pd

data = pd.read_csv('data.csv')
```

%% @insert-code-chunk plot data

++

```python exec
import matplotlib.pyplot as plt

# Assuming 'data' is a pandas DataFrame with 'year' as the x-axis and 'height' and 'width' as values.
# Plotting 'year' vs 'height'
plt.figure(figsize=(10, 5))
plt.plot(data['year'], data[' height'], label='Height', marker='o')

# Plotting 'year' vs 'width'
plt.plot(data['year'], data[' width'], label='Width', marker='o')

plt.xlabel('Year')
plt.ylabel('Measurements')
plt.title('Year vs Measurements')
plt.legend()
plt.grid(True)
plt.show()
```

++
````

### CODE REVIEW

**Comment by nokome** [comment]
Some further details to help with development...

The main task for this PR is to use Jinja to expose the `Context` to assistants in the system prompt template. I have very quickly added an example of how this could be done in the `assistants/builtin/insert-code-expression.md` file e.g. 

````markdown

# Languages

The following language runtimes are available.

{% for kernel in context.kernels %} 
## {{ kernel.info.name }} {{ kernel.info.softwareVersion }}

Operating system: {{ kernel.info.operatingSystem }}

### Packages

These {{ kernel.info.name }} packages are available:

{% for package in kernel.packages %}
- {{ package.name }} {{ package.version }}
{% endfor %}

### Variables

These variables are defined in {{ kernel.info.name }}:

{% for variable in kernel.variables %} 
Name: {{ variable.name }}
Type: {{ variable.nativeType }}
Structure: {{ variable.hint|tojson(true) }}
{% endfor %}

{% endfor %}

# Examples

Examples of code within the document:

{% for chunk in context.code_chunks %}
```{{ chunk.programmingLanguage }}
{{ chunk.code }}
```
{% endfor %}

````

Some things to note:

- Add `debug: echo` to the header of a specialized assistant to echo the rendered system prompt into the returned suggestion. This is very useful for begin able to see what the general assistant that is delegated to will actually receive. e.g. 

```markdown
### Variables

These variables are defined in python:

 
Name: __builtins__
Type: dict
Structure: null
 
Name: pd
Type: module
Structure: {
  "type": "Unknown"
}
 
Name: data
Type: DataFrame
Structure: {
  "columns": [
    {
      "itemType": "Integer",
      "maximum": 1992,
      "minimum": 1990,
      "name": "year",
      "nulls": 0,
      "type": "DatatableColumnHint"
    },
    {
      "itemType": "Number",
      "maximum": 2.3,
      "minimum": 1.2,
      "name": " height",
      "nulls": 0,
      "type": "DatatableColumnHint"
    },
 ```

- Note camelCased property names in context

- The `|tojson(true)` filter is useful for quickly dumping part of the context so you can see it in the echoed prompt

- Currently `tojson` is used for the variable `hint` but it feels like if would be better to use Jinja `{% if variable.type == 'DatatableHint' %}` etc to expand this into words.

PR #2109, opened 2024-03-07
fix(config): design updates to input fields and layout

### LINKED ISSUE #2097: Frontend Keys Design Audit
Things to consider:

- The windows need pop out a little more (colour/contrast wise)
- Turn off spellcheck within input fields
- Save or "Save keys" instead "Save me" 
- Clear field icon/button
- Not actually saving. - save button sends a get request.

Potentially look at having each field save itself?
- When you click out of the field it saves the request

simonwinter:
This PR addresses issues raised in #2097 

- The windows need pop out a little more (colour/contrast wise)
- Turn off spellcheck within input fields
- Save or "Save keys" instead "Save me"
- Clear field icon/button

I think I have sorted:

> When you click out of the field it saves the request

The key modification endpoint needs to be looked at:

> Not actually saving. - save button sends a get request.
> Potentially look at having each field save itself?

closes #2097


nokome:
I've been debugging the save value and I see too problems with the `POST` requests to set a key:

1. `value=` is being prepended to the body, it should just be the value, in this case `dddddddddddddddddd`
2. `Content-Length: 0` and have the confirmed that on the Rust server is not getting any body content (but does when I curl to that endpoint).

image

This is explains why they come back as just all dots when redacted - because the secret is set to an empty string so it just gets prefix with 12 dots.

nokome:
I've made several commits to fix the setting of secrets. The main one was moving the body out of the headers in the `POST` request: https://github.com/stencila/stencila/pull/2109/commits/8d0354ef6831af1a0118f2d0a486cfaaf0f56a2b. In addition:

- explicitly delete a secret if it is blank
- on the server if a client attempts to set a blank secret, delete it
- improve the redaction algo so it reflects the length of the key

Two issues remain:

- `GET ~/secrets` is fetched on every page load. It should only occur when the secrets config panel is opened

- For security reasons the server only ever provides redacted secrets. The problem is that the redacted value becomes the input value which the user can then modify and set on the server with ● characters. Users should not be able to modify existing keys, they should be able to paste in (or type maybe as well) new keys and delete existing keys. i.e CRUD without the U.

Generally I feel like this has been approached as a config setting solution. It feels like there is a lot of state management going on in the client, including detecting modifications, that may not be necessary. An alternative approach would be to treat operations on keys a lot more atomically and just fire a POST when a secret is set and a DELETE when the clear button is pressed.

nokome (approved):

PR #2128, opened 2024-03-14
Fix: execution ended datetime representations 2125

### LINKED ISSUE #2125: fix(`<stencila-ui-node-execution-ended>`): refresh the humanized representation and show the ISO 8601 date/time in the tooltip
The current implementation is here:

https://github.com/stencila/stencila/blob/bbe35ecb008cb630c0f7249eb70d09fee7badb34/web/src/ui/nodes/properties/execution-ended.ts

It is fine, but the humanized time does not update and the tooltip currently shows the raw timestamp.

@mike-parkin could you please work on this and submit a PR for it. I can't imagine it would take more than an hour but please let me know if you think it will because I might de-prioritize it in that case.

mike-parkin:
**details**

changes in the 'stencila-ui-node-execution-ended' element:
 - set the tooltip to display the time ISO like format: YYYY-MM-DDTHH:mm:ss.
 - start an interval in the connectedCallback() which updates the relatvie time display every minute.

**related issues**: closes #2125


nokome:
Thanks @mike-parkin. This had a lot of changes already merged into master so I just cherry pick the last, relevant commit into `main`.

PR #2136, opened 2024-03-15
feat(authors): update authors layout

### LINKED ISSUE #2132: fix(node authors): improve rendering of `<stencila-ui-node-authors>` as per designs
## Background

I've done some refactoring of the rendering of the `authors` property. The schema is flexible (attempting to stay consistent with schema.org while also allowing for more detail that we require), but this makes consistently rendering authors as per the designs complicated. To solve this I have created `<stencila-ui-node-author>` which has some rationale for its existence:

https://github.com/stencila/stencila/blob/493b71c58051cb030776e27cfd3e6cb197e133ec/web/src/ui/nodes/properties/author.ts#L7-L25

But currently very little styling (I just wanted to check all the necessary props were getting passed down):

image

Whereas the designs are like this:

image

To avoid bloating documents too much the author information for each node is kept minimal. It is envisaged that when we will store more complete user information for each author in the `authors` property of the `Article` and use `id` to fetch that to add things like affiliations. Maybe. Anyway, for now, there is limited info available.

## Tasks

- [x] I recommend using http://localhost:9000/nodes/paragraph/paragraph.json?view=split as a test - it has various permutations of the different author types.
- [x] Style `<stencila-ui-node-author>` and `<stencila-ui-node-authors>` as per the designs noting that not all aspects of the design will be possible (e.g specific avatars for people)
- [ ] Factor out most of `<stencila-ui-node-execution-duration>` into `<stencila-ui-node-timestamp-property>` in `ui/nodes/properties/generic/` so we can reuse it's updating humanized time for execution duration and for the timestamp associated with each author.
- [ ] There shouldn't be a need to touch the files in `web/src/nodes`, only those in `web/src/ui/nodes`. However, please review `AuthorRole.getAuthor()` - it seems to work but feels a little brittle and could do maybe do with some minor refactoring (e.g. factoring out the multiple `inner.tagName.toLowerCase()`).

simonwinter:
**Description**

The authors now renders as in the designs, including collapsing.

- Avatar based on role:
  - Org - building
  - SoftwareApplication, use the id to find icon
  - Person - Initial for now
  - stencila/insert - sparkle
  - stencila/edit - pencil
- Moved author to its own folder so I can include a utility file.
- Have finally switched to `Inter`
- As with other components, needs testing with other colour values
- Experimented with colour values, but it needs research to avoid contrast issues.
- set up a timestamp ui generic property. It simply transforms a timestamp via moment's fromNow function. I haven't merged the functionality from the execution ended component - the rendering is quite different. 

I haven't yet completed:

- [x] Factor out most of <stencila-ui-node-execution-duration> into <stencila-ui-node-timestamp-property> in ui/nodes/properties/generic/ so we can reuse it's updating humanized time for execution duration and for the timestamp associated with each author.

_\* I think this needs more review of the rendering use-case. The execution-duration etc include icons & tool tips. We should add the tooltip to the Author's date - but the message will need to be different_

- [x] There shouldn't be a need to touch the files in web/src/nodes, only those in web/src/ui/nodes. However, please review AuthorRole.getAuthor() - it seems to work but feels a little brittle and could do maybe do with some minor refactoring (e.g. factoring out the multiple inner.tagName.toLowerCase()).

_\* It appears to be working well for me - however, you're right it needs a little refactoring_

closes #2132


rust/server/src/documents.rs
```diff
@@ -361,7 +361,7 @@ pub async fn serve_path(
         format!(
             r#" <link rel="preconnect" href="https://fonts.googleapis.com">
                 <link rel="preconnect" href="https://fonts.gstatic.com" crossorigin>
-                <link href="https://fonts.googleapis.com/css2?family=Lato:wght@400;500;900&family=Montserrat:wght@400;600&display=swap" rel="stylesheet">
+                <link href="https://fonts.googleapis.com/css2?family=Lato:wght@400;500;900&family=Inter:slnt,wght@-10..0,100..900&family=Montserrat:wght@400;600&display=swap" rel="stylesheet">
```
nokome, line 364: Should we remove Lato, also Monserrat?


```suggestion
                <link href="https://fonts.googleapis.com/css2?family=Inter:slnt,wght@-10..0,100..900&family=Montserrat:wght@400;600&display=swap" rel="stylesheet">
```

web/package.json
```diff
@@ -96,6 +96,7 @@
     "prosemirror-state": "1.4.3",
     "prosemirror-view": "1.33.1",
     "split.js": "1.6.5",
+    "string-to-color": "2.2.2",
```
nokome, line 99: If this is not going to be used shall we remove it?

web/src/ui/nodes/properties/author/index.ts
```diff
@@ -0,0 +1,183 @@
...
+   */
+  private splitID() {
+    return this.$id?.trim().split('/') ?? []
+  }
+
+  private getInitial() {
+    return this.name.charAt(0)
+  }
+
+  private getSoftwareIcon() {
```
nokome, line 128: A bit picky, but I wonder if all these functions that `return html` should be prefixed with `render` instrad of `get`

web/src/ui/nodes/collapsible-card.ts
```diff
@@ -0,0 +1,53 @@
...
+import './properties/generic/collapsible'
+
+/**
+ * UI Node Collapsible Details
+ *
+ * A component to render inside a node-card, which allows its content to be
+ * collapsed & hidden. This includes the necessary styling of the header &
+ * the shell of the body.
+ */
+@customElement('stencila-ui-node-collapsible-details')
```
nokome, line 15: I'm don't fully understand why this is needed in addition to the existing `stencila-ui-node-collapsible-property`. But regardless, for consistency, this file should be move to `web/src/ui/nodes/properties/generic` folder and maybe renamed `collapsible-details.ts`.

web/package.json
nokome: Done

web/src/ui/nodes/properties/author/index.ts
nokome: Done

web/src/ui/nodes/collapsible-card.ts
nokome: Done

nokome:
I've addressed remaining todo items and comments and merged.

PR #2148, opened 2024-03-18
feat/pop-out-node-card

### LINKED ISSUE #2144: feat(pop-out node cards): implement a pattern for users to view node cards on demand
## Problem

There are various views for a document which vary in the amount of information and interactions that they expose to a user. For the "dynamic view" (user can inspect code and re-execute the document but not modify it) and "visual view" (user can do what they can in dynamic view plus edit the document) we have moved the designs towards having node cards "in-flow" rather than off to the side. This is necessary for the node types that have nested content (e.g. `InstructionBlock`, `ForBlock` and `IfBlock`) but also for node cards that have outputs that become part of the content (e.g. `CodeChunk` and `CodeExpression`). However there are node types where we don't want node cards to be always visible. These fall into two groups.

### Non-executable block node types

Currently, in the dynamic view, for non-executable block node types such as paragraph (and when implement heading, list, table etc) the node card is always shown. This is unnecessary and confusing. e.g.

image

### Non-executable and executable inline node types

Node cards are `display: block` and currently always shown. This is not appropriate for node types like `CodeExpression` and `MathInline` which have a lot of extra properties to display but where the node card breaks up the inline content:

image

## Solution

To avoid the above, the design proposed in https://github.com/stencila/stencila/issues/2027#issuecomment-1977838235 is to have an icon for the node type pop up on hover which allows the node card to be shown, interacted with and then hidden. Much like a "modal":

image

## Tasks

Implement the above display interaction for node cards via an additional attribute (`display`?) which can be specified in the node type specific components in `web/src/nodes` e.g.

```ts
override renderDynamicView() {
    return html`<stencila-ui-node-card type="CodeExpression" view="dynamic" display="ondemand">
      <span slot="header-right">
        <stencila-ui-node-execution-actions></stencila-ui-node-execution-actions>
      </span>
      <div slot="body">
```

simonwinter:
**Description**

Node cards can now be toggled via a button which is revealed when hovering over the node in the dynamic view.

e.g. 

https://github.com/stencila/stencila/assets/1104537/bd43fce7-9b93-467b-8412-2ee222dc8fca

---

**Implementation notes**

- Placing 2 slots inside the node card causing a number of styling issues. I've had to add a number of wrapper elements to make toggling the needed classes on/off easier.
- I've introduced an instance variable `ui` which captures the theming for the node via the `connectedCallback` lifecycle method.
- The max-height animation on the collapsible component does not work successfully if we are reactively toggling tailwind classes. Each time the toggle state is set, the entire css wrapped inside a tw `apply` call gets rewritten. To solve this I have introduced a `wrapper` component `ui/animation/collapsible`. This solves the issue by having styles set on the `:host` & allowing for a `.opened` class to be added/removed.
- The above is an issue in React and other frameworks.
- In a pure html/css world, the css should be simple:
```css
div {
  overflow: hidden;
  opacity: 0;
  max-height: 0;
  transition:
    max-height 400ms cubic-bezier(0, 1, 0, 1),
    opacity 400ms;
  transform: translate3d(0, 0, 0);
}

:host(.opened) div {
  max-height: 1000px;
  opacity: 1;
  transition:
    max-height 700ms ease-in-out,
    opacity 400ms;
}
```
- I think we will need to review the `collapsible` node card elements as they suffer the same issue.
- Now that I've set this up, the code is somewhat bloated. I'd like to refactor this into 3 or 4 sub components to make dealing with the layout + css a lot easier.

---

closes #2144


nokome:
Thanks for getting this to this stage @simonwinter 👏🏽  it is looking great. There are places that is breaks things e.g. `Paragraph`s inside list items and in the `source` view but lets come back to that later in the week.

PR #2156, opened 2024-03-21
fix(node-card): changed layout of details to an inline element

### LINKED ISSUE #2155: feat(node-card details): turn details into an inline display
### Summary

Details takes up too much space - could the items inside the card be displayed in one line?

simonwinter:
**Details:**

- Removed collapsing from the element
- Changed layout to an 'inline' element
- Reduced space & font size
- Added borders to separate this element from others
- Added a container query to change flex direction when container is too small

closes #2155


nokome (approved):
LGTM, thanks!

PR #2169, opened 2024-03-26
refactor(assistants): id->name + name->title

brettc:
Rename Assistant properties for consistency with Kernels.


nokome:
- as well as consistency with `Kernel::name()` this also provides consistency with `Codec::name()`.
- had to fix an erroneous rename of `mistral::Model.id`
- also renamed to `name` and `title` for `PluginAssistant`
- also renamed plugin assistant method to `assistant_perform_task`
- added `Assistant::availability()` as for kernels

PR #2173, opened 2024-03-27
feat(node card): reduce padding & vertical heights of node-card

### LINKED ISSUE #2154: feat(node card padding): reduce vertical height of node cards
### Summary

Look at vertical padding around the elements in the node cards

simonwinter:
**Description**

Resolves:

* Some vertical padding of node-card elements
* Reduces font sizes in card elements
* Fixes\* missing vertical scrolling

\* This appears to be resolved, but I can't be 💯 certain. It really needs some more testing.

We will need to come back & look at this in more detail. There are issues such as elements that have no content rendering on screen & taking up vertical space in the content flow. 

closes #2154


nokome (approved):
Thanks @simonwinter, will merge now

Issue #2187, opened 2024-04-03
feat(node-card): Rendering issues to address


simonwinter:
### Summary

Following on the work of #2174, there are a few issues that have fallen out of this issue:

- [x] there are new `in-flow` & `on-demand` components. They are copy+pasted from the original `node-card` component. There's much that can be shared, particularly the styles between the components. Let's see what we can rationalise here.
- [ ] The `styled-block` and `styled-inline` components are causing some rendering issues (see below). Ideally we'd be able to merge the styles together to resolve these issues.
- [x] Output overflow not providing scrolling.
- [ ] Tooltips displayed on the `on-demand` instances are not displaying correctly (i.e. missing bg colour)
- [ ] If a node-card's title is a programming language, ensure that it renders the correct title, capitalisation etc.

![CleanShot 2024-04-03 at 16 55 26](https://github.com/stencila/stencila/assets/1104537/33e60501-b4a8-4b02-aa45-01bcd5b71255)
_Styled block not merging bg colour & padding. Also note the elements are not displayed full width_

![CleanShot 2024-04-03 at 16 57 44](https://github.com/stencila/stencila/assets/1104537/8c8db250-a086-4256-b5a4-c8a94075534e)
_Inline block not merging bg colour and padding_

![CleanShot 2024-04-03 at 16 58 55](https://github.com/stencila/stencila/assets/1104537/fa9eed09-1ebf-4b18-b956-cf15140bfa15)
_Dynamically generated classes are not rendering as expected_

![CleanShot 2024-04-03 at 17 00 07](https://github.com/stencila/stencila/assets/1104537/e5cd6e9b-98bf-4561-a663-0f24a145b2e3)
_Missing scrollbar on output with overflow_

![CleanShot 2024-04-08 at 13 57 27](https://github.com/stencila/stencila/assets/1104537/3879f58f-6f10-46cf-bffe-392f3d495655)
_Tooltips on top of tooltips!_

nokome:
Closing as mostly done

PR #2188, opened 2024-04-03
Inline/block on demand & in flow

### LINKED ISSUE #2174: feat(node-card): components to render inline/block for on-demand and in-flow
### Summary

**Description**

There are 3 use-cases for rendering elements that we need to handle:

_**on-demand**_

- A `block` level element: render as a block, users can click on the `chip` which renders above the element
- An `inline` level element: render inline, clicking on chip renders details _above_ the parent element

_**in-flow**_

- A `block` level element which include the node card rendered without the need to have a user enable/disable it.

simonwinter:
**Description**

Previously we had a `node-card` component which allowed us to render a node in either `on-demand` or `auto` modes. `on-demand` was for allowing blocks to be rendered as users interacted with them, `auto` - the element rendered on the screen as is.

This did not take `inline` vs `block` elements into account.

This PR provides some new components:

- `stencila-ui-block-in-flow`: `ui/in-flow/block.ts`
- `stencila-ui-block-on-demand`: `ui/on-demand/block.ts`
- `stencila-ui-inline-on-demand`: `ui/on-demand/in-line.ts`

These have been applied to the following (plus some other minor node types):

- `CodeChunk`
- `CodeExpression`
- `CodeBlock`
- `CodeInline`
- `MathBlock`
- `MathInline`
- `StyledBlock`
- `StyledInline`

_Note:_ Some rendering issues have fallen out of this issue. See #2187 for details on next steps.

---

closes #2174

### DIFF
--- a/web/src/nodes/array.ts
+++ b/web/src/nodes/array.ts
@@ -8,6 +8,7 @@ import { withTwind } from '../twind'
 import { Entity } from './entity'
 
 import './array-item'
+import '../ui/nodes/in-flow/block'
 
 /**
  * Web component representing a Stencila Schema `Array` node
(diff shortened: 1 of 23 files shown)


nokome:
Sorry, I had forgotten to merge this!

PR #2192, opened 2024-04-09
feat(quickjs kernel): add a QuickJS kernel and make default

nokome:
Currently the Rhai kernel is used as the default kernel for executing and evaluating code that does not have a specified language. Rhai was chosen because it is embed-able and so does not rely on an external process. This PR adds a QuickJS kernel which is also embed-able but uses a language that is far more widely known.


nokome:
Merged

PR #2204, opened 2024-05-01
Feature/on demand node cards

mike-parkin:
**details**

Have replaced the 'in-flow' block node cards with 'on demand' cards for the dynamic view, and inserted the output from the nodes and into the 'content' slot of card.

image

image

There are still some inherited styles such as the width of the content, which are going to be resolved in another working branch.

**related issues**: N/A


rust/server/src/documents.rs
```diff
@@ -355,6 +355,7 @@ pub async fn serve_path(
                 <link href="https://fonts.googleapis.com/css2?family=Inter:slnt,wght@-10..0,100..900&family=IBM+Plex+Mono:wght@400&display=swap" rel="stylesheet">
                 <link rel="stylesheet" type="text/css" href="/~static/{version}/shoelace-style/themes/light.css">
                 <link rel="stylesheet" type="text/css" href="/~static/{version}/shoelace-style/themes/dark.css">
+                <link rel="stylesheet" type="text/css" href="/~static/{version}/shoelace.css">
```
nokome, line 358: I don think this is necessary because in `web/src/views/dynamic.ts` we have 

```ts
import '../shoelace.css'
```

and so Parcel bundles that CSS into `web/dist/views/dynamic.css`. Adding this here doesn't really hurt too much but does adds another unnecessary request and more importantly leaks the frontend architecture into the server. Id prefer if the server just served up the CSS and JS for the relevant view and the front end is responsible for putting what is necessary in those.

Can you please check if this change is necessary.

web/package.json
```diff
@@ -22,6 +22,7 @@
   ],
   "type": "module",
   "source": [
+    "src/shoelace.css",
```
nokome, line 25: Related to previous comment, remove if not necessary.

web/src/nodes/instruction-inline.ts
```diff
@@ -1,10 +1,16 @@
 import { NodeType } from '@stencila/types'
+import { html } from 'lit'
 import { customElement } from 'lit/decorators.js'
 
 import { withTwind } from '../twind'
 
 import { Instruction } from './instruction'
 
+import '../ui/nodes/node-card/on-demand/in-line'
+import '../ui/nodes/node-card/in-flow/block'
```
nokome, line 10: Is this import necessary?

web/src/nodes/math-block.ts
```diff
@@ -25,7 +25,7 @@ export class MathBlock extends Math {
    */
   override renderDynamicView() {
     return html`
-      <stencila-ui-block-in-flow type="MathBlock" view="source">
+      <stencila-ui-block-on-demand type="MathBlock" view="source">
```
nokome, line 28:
```suggestion
      <stencila-ui-block-on-demand type="MathBlock" view="dynamic">
```

nokome (changes requested):
Thanks @mike-parkin - left a few minor comments.

PR #2207, opened 2024-05-06
Fix/node themes

simonwinter:
**Description:**

This updates the default theme to render both documents (static view) & the web view in _vscode_ (dynamic view). The changes here ensure:

- We have content restricted to the container dimensions (tables, images, figures etc)
- There _shouldn't_ be situations where the parent element affects the rendering of node-cards (i.e. font weights should not be inherited)
- Introduces tables, figures etc so that they are also rendered as collaped.

The work here also includes work by @mike-parkin in `feat/table-node-card` and `feat/code-chunk-outputs`

**Known issues:**

In order for tables to have an overflow scroll & allow for hovering with the node card, tables appear "indented". This gives room on the left hand side for the first column of elements to be hoverable & have their "cards" discovered.

CleanShot 2024-05-06 at 17 43 08

Currently in the dynamic view, due to rendering requirements of the node cards, list items appear to not be aligned. I believe this is due to how the max height of the animation component is being applied. I think this is solvable.

CleanShot 2024-05-06 at 17 52 34

**Missing static rendering?**

@nokome - do you have time to look at this issue? It's a blocker for merging this PR.

When testing the default theme, I've found a number of instances where the static view doesn't get the content rendered as happens in the dynamic view.

As seen on:

http://localhost:9000/examples/kernels/code-chunk/output-types-r?view=static
http://localhost:9000/examples/kernels/code-chunk/output-types-r?view=dynamic

| static | dynamic |
| ------ | ------- |
| CleanShot 2024-05-07 at 14 07 08 | CleanShot 2024-05-07 at 14 09 15 |
| CleanShot 2024-05-07 at 14 10 27 | CleanShot 2024-05-07 at 14 11 00 |
| CleanShot 2024-05-07 at 14 12 46 | CleanShot 2024-05-07 at 14 13 21 |

In the case of the datatable, we can see that there is a staticRender method, which doesn't appear to be getting called at all. see: https://github.com/stencila/stencila/blob/b6a5a45e5097b50b6e0f9151bc2bfa39c229eaf5/web/src/nodes/datatable.ts#L22-L26


nokome:
> When testing the default theme, I've found a number of instances where the static view doesn't get the content rendered as happens in the dynamic view.

What is happening here is that when the HTML is generated in Rust some things that are normally considered as content are being encoded as attributes. This is often done to make things easier for WebComponents, in particular reactive updates e.g.

```html
        <stencila-code-inline id=xxx code='2 * pi * r^2'>
        </stencila-code-inline>
 ```
 
This should be encoded as:

```html
        <stencila-code-inline id=xxx code='2 * pi * r^2'>
            <code>2 * pi * r^2</code>
        </stencila-code-inline>
 ```
 
 I can make that change for most of the ones you have pointed out above. I think that is best done as a separate PR so you can more easily check that the content I add does not end up getting "duplicated" in the rendering.

Re `renderStaticView` not getting called. I'm confused now as to why we/I thought that we needed to have WebComponent rendering for the static view - which should really be Javascript free (and are currently served without JS (?)). It may have been to do with math or styled nodes but in anycase that should not be necessary. I think we should have another separate PR where we nuke the `views/static.ts` and related methods  - it would be less confusing.

nokome (approved):
This is looking really great Simon! I personally like having all the nodes on demand and "discoverable" (although I do think we need a top level toggle to be able to easily show/hide all chips and show/hide all cards).

There are a bunch of refinements that can be made to styling and UX but I suggest that we leave that for a collective audit perhaps at the end of this week.

PR #2208, opened 2024-05-06
Feat: code chunk outputs & error messages

mike-parkin:
**details**

A refactor of code chunk properties

- refactor all the output types (array, obejct, bool, str, num, int, dt) to on-demand blocks so only the values will show in flow.
- show the exe messages in a linting extension of the codemirror
- remove the old exe messages section, and remove the collapsible header of the 'code' section

image


web/src/ui/nodes/properties/code/utils.ts
```diff
@@ -0,0 +1,59 @@
...
+import { ExecutionMessage } from '../execution-message'
+
+/**
+ * Convert the `ExecutionMessage.level` value into a @codemirror/lint `Severity` string
+ * for the linting extension
+ * @param lvl `MessageLevel`
+ * @returns 'Severity'
+ */
+const getMessageSeverity = (lvl: MessageLevel) =>
+  lvl === 'Error'
```
nokome, line 15:
```suggestion
  lvl === 'Error' || lvl === "Exception"
```

nokome: Should treat an `Exception` message the same as an `Error`.

nokome (changes requested):
Thanks @mike-parkin : one small but important change needed please.

nokome:
@mike-parkin There are now merge conflicts on this branch after I merged in another one. Can you please fix those and merge.

PR #2213, opened 2024-05-08
feat(myst plugin): add plugin for MyST

nokome:
Currently just initial setup.


rowanc1:
👀🚀

Excited about this, let me know when to jump in!

nokome:
@rowanc1 You're help would be awesome! I'm going to merge this now and @vijay-prema is going to make a start fleshing it out over the next could of weeks.

plugins/myst/package.json
```diff
@@ -22,6 +23,7 @@
   "dependencies": {
     "@stencila/plugin": "^2.0.0-alpha.1",
     "@stencila/types": "^2.0.0-alpha.25",
+    "mdast": "^3.0.0",
```
rowanc1, line 26: We also have `myst-spec` and `myst-spec-ext` which is our staging ground for types. Those two libraries are the main ones that are defining our mdast-extended types.

plugins/myst/src/decode/inlines.ts
```diff
@@ -0,0 +1,75 @@
...
+    switch (md.type) {
+      case "text":
+        return new Text(md.value);
+      case "emphasis":
+        return new Emphasis(mdsToInlines(md.children));
+      case "strong":
+        return new Strong(mdsToInlines(md.children));
+      case "inlineCode":
+        return new CodeInline(md.value);
+      case "mystRole":
```
rowanc1, line 46: Depending on how you are implementing the `mystRole`/`mystDirective` is mostly a parsing container, we generally get rid of these by the time we render / translate to other formats. For example here, we remove the `sup` role, and just use the children in a paragraph.

[image](https://mystmd.org/sandbox?tab=ast&myst=ewBzAHUAcAB9AGAAYQBzAGQAZgBgAA%3D%3D)

plugins/myst/src/snapshots/marks.json
```diff
@@ -0,0 +1,61 @@
+{
+  "type": "Article",
+  "content": [
+    {
+      "type": "Paragraph",
```
rowanc1, line 5: Our types are lower case (and camelCase for things like `inlineCode`), maybe these are stencilas types?

rowanc1 (commented):
Nice - looks like you are up to speed with the `myst-parser`, left a few thoughts as I looked through it!

plugins/myst/src/decode/inlines.ts
nokome: Yep, I think we are on the same page here, this is just a `case` to unwrap the `mystRole`.

plugins/myst/src/snapshots/marks.json
nokome: Yes, these are snapshots of Stencila Schema types transformed from MyST types.

plugins/myst/src/decode/inlines.ts
rowanc1: There are a lot of transformations that we run on the AST, the most basic of which [are here](https://github.com/executablebooks/mystmd/blob/main/packages/myst-transforms/src/basic.ts#L18). This one [specifically lifts the roles/directives](https://github.com/executablebooks/mystmd/blob/feat/remote-url/packages/myst-transforms/src/basic.ts#L20).

Issue #2206, opened 2024-05-06
feat(provenance): display provenance in preview


nokome:
## Background

We already have an `authors` property on many node types (and populate that for machine generated content; doing the same for human generated content via an editor in the works).

Now, in addition several node types have a `provenance` property. This represents a summary of who (machine or human) did what (wrote, edited, verified) to the current content in the document.

This is all recorded deep in the bowels of Stencila's patching algorithm and is made efficient using [run length encoding](https://en.wikipedia.org/wiki/Run-length_encoding). Each run records the total count of authors, the last eight authors and a "provenance byte" [(which encodes about as much provenance as you can in a single byte)](https://github.com/stencila/stencila/blob/039a3246056fb983594b909d2352915184a8b58b/rust/schema/src/cord_provenance.rs).

The following is HTML output from running the following on the new example file:

```sh
cargo run -p cli convert examples/provenance/simple-1.json --to dom --pretty
```

Nodes such as `Paragraph`, `Section` and `Article` has a `provenance` field that is a vector of `ProvenanceCount` that gives the character count and percent by provenance category within that node.

The VSCode extension presents this information in a CodeLens:

![image](https://github.com/stencila/stencila/assets/1152336/62e45c2a-3097-4e58-a8bb-8ac4a285ee4c)

## Displaying provenance summaries in node cards

- [x] A. The `<div slot=provenance>` should be displayed in node cards (as a single line above or below authors perhaps? with tooltips explaining each category?). There are many different [categories](https://github.com/stencila/stencila/blob/main/schema/ProvenanceCategory.yaml) but any one node is only likely to have 2-5. Search `schema/*.yaml` for `provenance` to see which nodes have it - all those with `authors` I believe. 

## Displaying authorship runs in text

The `<stencila-authorship>` elements within `<stencila-text>` should be displayed:

- [x] B. Different shades of blue for different provenance (note this is a string representation of the provenance byte and is similar to, but not the same as, the `ProvenanceCategory` enumeration; see [tests](https://github.com/stencila/stencila/blob/039a3246056fb983594b909d2352915184a8b58b/rust/schema/tests/cord_provenance.rs#L4) for examples of possible values e.g `MwHeHv3Mv`). Take inspiration from the groups in `ProvenanceCategory.yaml` (linked to above) for how to shade. This shading should only be enabled when the node card is shown (long term probably also want a global way of turning this on?).

- [x] C. Tooltips for each `<stencila-authorship>` run showing the `count` and `provenance`

[...]

PR #2214, opened 2024-05-08
Feature: stencila authorship component 2206

mike-parkin:
**details**

web-compoent for `<stencila-authorship>` created
- adds the background shading based on the 'mi' property.
- tooltip on hover, shows author count and provenance.
- only shows the provenance when the relevant `Entity`'s node card is open

image
image

Due to complexity issues, the current toggling of the provenance can only be attached to a single `Entity`, 
__for example:  a Paragraph entity will only display its provenance shading when the Paragraph node card is open, not when the Instruction block is open at this point__

This is due to the `CustomEvents` that control the context values bubbling up and triggering multiple listeners.
Expanding 

**related issues**:  #2206 items B & D


web/src/ui/nodes/icons-and-colours.ts
```diff
@@ -151,3 +151,22 @@ export const executionMessageUI = (
...
+export const provenanceHighlights = {
+  0: 'transparent',
+  1: '#f1f5fe', // blue-50
+  2: '#cfdefc', // blue-100
+  3: '#b1c9fa', // blue-200
+  4: '#92b3f7', // blue-300
+  5: '#77a0f5', // blue-400
+}
+
+export type ProvenanceHighlightLvl = keyof typeof provenanceHighlights
```
nokome, line 166: You and your cutbacks! To save an old fart like me a few brain cycles could your neo-liberal fingers spare a couple more characters?
```suggestion
export type ProvenanceHighlightLevel = keyof typeof provenanceHighlights
```

mike-parkin: I'm surporised you didn't comment on this one: 
```
export const getProvHighlight = (miLvl: ProvenanceHighlightLvl) =>
```

nokome: Neuseating

PR #2216, opened 2024-05-08
Feature/provenance slot on node cards

simonwinter:
**Description**

This addresses todo item `A` in issue #2206 - displaying provenance on node cards.

The `stencila-ui-node-provenance` card includes a single slot, which is expected to be populated with contents from a `<slot name="provenance">` slot.

```ts
<stencila-ui-node-provenance type="Paragraph">
    <slot name="provenance"></slot>
</stencila-ui-node-provenance>
```

The component will look for `stencila-provenance-count` children passed into the slot & look for `provenance-category` properties. 

> [!WARNING]
> This is not great for build - there must be a way to get the descriptions back from the server? Even adding a `provenance-description` to each `stencila-provenance-count` component, though repetitive could be better?

It then uses these keys to look up the the corresponding description in the schema (via a symlinked schema json).

https://github.com/stencila/stencila/blob/fdc058816cab98accdae542f478cb274cac8bb04/web/src/ui/nodes/properties/provenance/provenance-category.ts#L10-L23

**Display**

Each node card has a new "provenance" property section directly below the authors properties. It re-uses the header from the collapsible sections, but instead of it being collapsible, it displays the categories as a series of "chips" to the right of the header. Each chip can be hovered to reveal the tooltip with the category description.

https://github.com/stencila/stencila/assets/1104537/1f680b0c-3e63-4555-b3c9-59e6f996f331

**Concerns/issues**

The icon is new - handshake seemed like a good metaphor, open to ideas. I think the placement below authors makes sense. The "chips" are a new design element ... I think they feel part of the same language. They take inspiration from the message property design.

CleanShot 2024-05-08 at 22 57 23


nokome:
Thanks @simonwinter, some quick comments:

I thought these chips would have percentages in them (taken from the `character-percent` attribute):

image

Regarding this:

> This is not great for build - there must be a way to get the descriptions back from the server?
> It then uses these keys to look up the the corresponding description in the schema (via a symlinked schema json).

Doing an import of the JSON seems fine to me (does it need to be symlinked?) not sure what the concern is. But I think the best approach would be to generate the TypeScript object you need when we generate all the other types - we did this before for something bespoke that Mike needed.

PR #2217, opened 2024-05-09
Feature/missing nodes

simonwinter:
**Description**

This PR adds some missing nodes as discussed in Discord: https://discord.com/channels/709952324356800523/1236894762611773490/1237615433901670420

Specifically, we add:

- `Claim`
- `QuoteBlock`
- `Section`

This merges in work done in #2216 

They are now set up similar to other existing blocks (i.e. are wrapped in on-demand instances).

I haven't looked at `Admonition` yet as it appears it would need further discussion to move it along.

**To do**

@nokome The server needs to allow for the `body` as well as `content` slots. Currently we are only receiving `content` and are therefore mossing out on the author & provenance information.


simonwinter:
@nokome I need to merge this to progress. I think this will still be an issue to address:

> @nokome The server needs to allow for the body as well as content slots. Currently we are only receiving content and are therefore mossing out on the author & provenance information.

PR #2221, opened 2024-05-13
Fix/provenance highlights

mike-parkin:
**details**

 - Revert the authorship tooltip back to simple black/white tooltip, and re style codemirror tooltip to look that same
 - drop the shading of provenance highlighting, have converted the hex codes into rgba values, and dropped the alpha value to dim the shade.
 - reinstate the icon for the code executable nodes


nokome:
Thanks @mike-parkin. Have checked locally and works as expected. Still getting unwanted gaps when node cards are expanded but will create an issue for that with screenshots.

PR #2226, opened 2024-05-14
feat(author-provenance): new rendering component in article

### LINKED ISSUE #2224: fix(document preview): improve display of article level `author` and `provenance` properties
Currently, when using the VSCode extension, the user is automatically recorded as an author of the paper which is rendered in HTML

image

which looks like this:

image

This could be improved, probably by making this information "on-demand"

### LINKED ISSUE #2225: fix(provenance): use icons for displaying provenance categories
### Summary

Currently, in the preview, the provenance categories are displayed use letters e.g.

image

In the VSCode Markdown editor we now show Mw as an icon:

image

We should have consistency, using robot, person and tick (for verified) icons in the preview. The following is (old) code that was being used in the language server for guidance:

image

simonwinter:
**Description**

This PR adds a new component to show/hide author & provenance information in an article.

Initially I looked at re-using on-demand blocks, however there were a few UI issues:

1. We need a title to allow the node-card's chip to attach to something
2. We'd have to somehow override the colours used - it's a different node type from the existing node-cards.
3. When the node-card is expanded, the title will display in the content are of the card

Basically, it wasn't looking that hot.

**Solution**

Instead, I've introduced a new component here. By default we see a label "Authors and provenance". On clicking this, we expand the panel & show the authors & provenance.

I wanted the chrome around this element to only be visible if you are interacting with the element, so if the element is open & you hover outside, the border fades out. I think this means users can read the details without having any UI elements break their reading flow.

CleanShot 2024-05-14 at 23 20 24@2x

CleanShot 2024-05-14 at 23 20 51@2x

CleanShot 2024-05-14 at 23 21 05@2x

**Side effects**

I wanted to have an up `chevron` icon to show the panel is collapsed. The existing chevron was not allowing for an up version. This always baffled me - collapsing to a left chevron does not map to what's happening in the UI. The other instances of the chevron should now use this instead.

**Issue?**

An issue I have seen (related to the provenance count) is where sometimes I see 2 provenance counts on the same enity. They are different instances (different `id` & `character-count` properties). I think this has to do with 1 entity (in this case paragraph) containing another entity (in this case CodeExpression). I'm not sure that someone looking at this understands the difference - perhaps at some point a hover would highlight the corresponding element?

CleanShot 2024-05-14 at 23 06 04@2x

closes #2224
closes #2225


nokome:
Thanks @simonwinter this looks really nice. A couple of points:

- For some reason, in VSCode the provenance pills are being stacked for me:

image

- I made some minor changes to allow for zero percent (which means <1%)

- I think it would be good to somehow combine the toggling of author info with the toggling of node chip and card visibility menu that is working on - one top level icon for all that "toggling visibility" stuff.

PR #2240, opened 2024-05-28
refactor(dom html): add label to captions of `Table`, `Figure`, and `CodeChunk`

nokome:
This add a `<span class="figure-label">` or a `<span class="table-label">` to the first paragraph of the `caption` (if any) of `Figure`, `Table` and `CodeChunk` nodes. This is done because doing this via CSS and/or CSS was proving difficult.

There are some other changes which make the generated HTML more semantically correct:

- for `Figure`, nest `<figcaption>` within `<figure>`
- for `Table`, nest `<caption>` within `<table>`
- for `CodeChunk`,  put the caption before the `outputs` if the label type is a table, and after if it is a figure

See the diffs for the HTML snapshots.

### DIFF
--- a/examples/nodes/article-ark/article-ark.dom.html
+++ b/examples/nodes/article-ark/article-ark.dom.html
@@ -252,33 +252,35 @@
       </div>
     </stencila-styled-block>
     <stencila-table id=xxx depth=1 ancestors=Article>
-      <table slot=rows>
-        <stencila-table-row id=xxx depth=2 ancestors=Article.Table row-type=HeaderRow>
-          <tr slot=cells>
-            <stencila-table-cell id=xxx depth=3 ancestors=Article.Table.TableRow>
-              <td slot=content>
-                <stencila-paragraph id=xxx depth=4 ancestors=Article.Table.TableRow.TableCell>
-                  <p slot=content>
-                    <stencila-text id=xxx depth=5 ancestors=Article.Table.TableRow.TableCell.Paragraph>A</stencila-text>
-                  </p>
-                </stencila-paragraph>
-              </td>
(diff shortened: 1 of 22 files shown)


mike-parkin:
Is it possible with the Table content to get the <caption> and the table element as siblings <table> and assign the [slot="rows"] property to the table? as the dev tools show, the <table> element is currently filling the content <slot/> in the `Table` node, and the slots inside are not accessibile.



currently it only works with this set up:

image

image

image

I would like to be able to structure the component like so:

image

this is mainly to keep the captions outside the scroll container for the table, because it looks a bit wierd:

image

Due to the way table elements display in html Its a bit more difficult to apply the scrolling rule to the <tbody>  without changing up the display properties of the table.

I would like to apply the same figure element, but this is less of an issue as the figures are not kept inside scrolling containers.

mike-parkin:
Thanks @nokome :smile:

mike-parkin:
This is ready for a look whenever you have time @nokome,

I've kept the caption styling quite simple, but just list any preferences for the default theme (if you have them) and I'll add those :)

nokome:
Thanks @mike-parkin. I noticed that the figure label (for normal figures) was inconsistent. e.g. in VSCode with

```
::: figure

Caption

![](https://source.unsplash.com/random/200x300)

:::
```

image

I tried to fix that myself but found the CSS a little daunting and no obvious fix. Can you fix it please. 

While you are at it, can the rules for `stencila-table` and `stencila-figure, stencila-code-chunk` be merged? I feel like that would be good for maintainability and consistency (e.g. if someone uses the `default.css` as the base for their own theme).

Issue #2236, opened 2024-05-20
Rendering / Expanding Node Cards


brettc:
This issue is for discussing how best to deal with the node-card expansion in the preview.

There are a few concerns here. 

* The icons are overlapping, and it is difficult to see them piled on each other.
* Secondly, the card animations are jarring when opened. There are two issues here. First, the icon disappears and gets inserted in the header. Additionally, the text is reflowed as it now occupies a narrower width, as we try to embed within the card.

It is worth separating out some things we want to do here:
1. Show the provenance / execution details etc for a particular node.
2. Show the *scope* of a particular node, i.e. that a Figure wraps both code and image and a caption.

The problem is compounded by the fact that some nodes are *invisible*: they do not contain any text that we will see in a final document (the commands for assistants). So it is difficult to put icons next to these!

Below are some starting ideas (half thought through)

brettc:
![CleanShot 2024-05-20 at 17 10 51@2x](https://github.com/stencila/stencila/assets/457283/067a5dd3-c4a9-4a08-8a08-558c3e1ee548)

We should avoid trying to embed the cards. Adding cards should just insert the cards into the vertical flow (visually, that is; they could still be inside the card, just without the left/right margins being added)

brettc:
Use the menu to have different "modes" where the invisible nodes take up a full line even when they are not expanded.
This will ensure that we don't get overlappping icons.

Note that the heading runs into the margin. 
![Drawing 3](https://github.com/stencila/stencila/assets/457283/274bdca7-271f-4950-afff-650d406e49bd)

brettc:
This solves some problems, but we do not get the scope shown as we did before. A few thoughts: 
* I think the embedding nodes were not doing a good job of showing scope anyway.
* The tree / outline view actually gives you a much better idea about that.

So, a few thoughts:
* We could try and sync the outline view.
* We could try and do something a bit fancy in the margins. But we are going to have to deal with some deep hierarchies in some cases, so this will required some thought.

brettc:
Another related comment. I think [vertical rythym](https://iamsteve.me/blog/a-guide-to-vertical-rhythm) really makes documents look better. To aid in this, here is a suggestion.

* We have only two line heights. Headers, and normal text.
* We distinguish headers of different level by prefixing H1 H2 H3 (and maybe font types).

In addition, I think vertical space is precious, and we are wasting too much of it. We remove much of the vertical whitespace in the node cards, by

* reducing lines height (as above) 
[...]

PR #2241, opened 2024-05-28
refactor: node chip behaviour

mike-parkin:
**details**

changes some simple positional and behavioural properties of the node card toggle chip

- Node chips remain visible when the card is expanded. and clicking again will collapse.
- chip icon is swapped to chevron when card is open.
- node chips are now positioned relative to the 'stencila-ui-collapsible-animation' component, instead of the content container, this keeps them inline with the top corner of the NodeCard. and also helps spread out the card chips.
- inline chips have be re-positioned, to avoid overlapping the inline card when it opens.

image

image

**related issue**:  #2236


nokome (approved):
Looks good, thanks @mike-parkin

PR #2247, opened 2024-05-30
Feat: use text opacity to display provenance

mike-parkin:
**details**

Have swapped the provenance highlight out for opacity of text
 -> has the 6 (0->5) mi levels from `1 -> 0.5` opacity

in the content <stencila-authorship> component, the opacity is set on the text only (via the `--tw-text-opacity` alpha variable)

for the provenance in the code view, I can only set the elements opacity, this is not ideal, but does have a side effect of varying opacity when the text is selected see image below.

As the text in the codemirror is colored by the language extensions, it would be virtualy impossible to apply the opacity to the text only (until css acutaly creates a `text-opacity` property)

image

### DIFF
--- a/web/src/ui/nodes/icons-and-colours.ts
+++ b/web/src/ui/nodes/icons-and-colours.ts
@@ -154,19 +154,19 @@ export const executionMessageUI = (
 
 // Provenance Highlight Colours ----------------------------
 
-export const provenanceHighlights = {
-  0: 'transparent',
-  1: 'rgb(241, 245, 254, 0.9)', // blue-50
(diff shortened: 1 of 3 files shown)


mike-parkin:
Oh! and an issue that will need to be addressed, if we go ahead with this: 
we will need to get the text color (if set via a style block) down to the authroship component, because in order to set the text opacity, we also need to set the `color` at the same level.

nokome:
Been testing this out. Some screenshots (mainly because your description made me think this wasn't going to be as good as it is!):

image

image

Issue #2251, opened 2024-06-09
fix(web): various improvements to web components and document preview


nokome:
- [x] Styling of inline nodes (e.g. `CodeInline`, `MathInline`, `CodeExpression`): work-in-progress
   - @mike-parkin we discussed me adding a `<code>` element to `CodeExpression`, however after looking at what will be required to do that on the Rust side (writing a custom HTML serialization method) I'd prefer not to do that and instead only should the `<output>`. If there is no output, then nothing will be visible, but that is OK for now.
   
- [ ] Node chip offsets
  - use the new `depth` attribute on `<stencila-*>` nodes to offset chips to the bottom right
  - use naive, basic offsets at first but consider adding a maximum right offset for depth > 3 or 4 to avoid overlaying node cards
 
- [x] Node card improvements
  - [x] Change caret in top-right to collapse the card `body` (currently it hides the whole card, which is provided by the chip)
  - [x] Move the provenance indicators (e.g. 🤖 100%) into the "Authors" heading
  - [x] Reduce and unify vertical space and padding in node cards
  - [x] Remove double borders within node card sections
  - [x] Change the font color in node cards to be a dark shade of the base card color e.g. dark blue for paragraphs (do this allowing for likelihood of at least some card colors changing)
  - [x] For executable nodes, improve the appearance of execution status when the not has not yet been executed
  - [x] For executable nodes, simplify the execution buttons to a single "Run" button with a `<sl-dropdown>`  for (Run above, Run below etc)
  - [x] For inline nodes, make the node card wider (currently constrained by the tooltip width?)
 
- [x] Document menu
  - Make this into a `<sl-menu>` and divide into sections (to be discussed)

- [ ] VSCode editor-preview linking @nokome 
  - [ ] #2310
  - [ ] #2311
  - [x] Clicking on a node's "View" code lens in editor should scroll the node into view (if not already) and expand the node card for that node

mike-parkin:
Thanks @nokome  :)

nokome:
A few more items:

- [ ] Color or executable "play" button
- [ ] Border under code in `CodeChunk` component
- [ ] For `CodeBlock` component do not show line numbers; in theme improve display (border?)

nokome:
Closing, remaining issues have been spawned into new.

PR #2252, opened 2024-06-09
Feat/restyle inline nodes

mike-parkin:
**details**

restyles the inline content of some inline nodes, removes the coloured wrapper around the inline node,

have added a style to the `CodeInline` content (inspired by the discord inline code with visible backticks)

`CodeExpression` now will only show the outputs in its content, so if there is no output, nothing will be visible, but the chip can still be activated.

also fixed the transparent unstyled tooltips that were appearing on the inline node cards exexution buttons.

**related issues**:  #2251


web/src/ui/nodes/node-card/on-demand/in-line.ts
```diff
@@ -72,15 +78,18 @@ export class UIInlineOnDemand extends ToggleChipMixin(UIBaseCard) {
...
+        color: ${(colors['black'] ?? 'black') as string};
+        padding: 0;
+        border-radius: 0;
+        background-color: transparent;
         pointer-events: all;
       }
 
+      &::part(base__arrow) {
+        display: none;
+      }
```
mike-parkin, line 91: I've refactored these styles to standard css properties/values, as setting the custom shoelace properties was cascading down to any instance of the `sl-tooltip` below this, eg the execution buttons.

PR #2253, opened 2024-06-10
Draft: feature - apply vertical offset to nested block node chips

mike-parkin:
**details**

Have started to set offsets to the node chips for block nodes,

so far have implemented a basic function which sets the 'Y' offset by 5px per depth level (after excluding)
Have left the x offset for now, as there is already a very small natural x offset for nested nodes anyway.

Current Issues:
 - There are some style issues with the X offset.
 - would like to park his whilst finalising the other changes to the node card, as these changes may affect some styles.

**related issue**:  #2251


mike-parkin:
have set Y offset to 6px

before

image

after

image

PR #2254, opened 2024-06-13
Feat/node card improvements 2251

mike-parkin:
**details**

This PR addresses all the node card related from #2251 , most are complete or almost complete

details for each below.

-----------------------------

- [ ] Node card improvements
  - [ ] Change caret in top-right to collapse the card `body` (currently it hides the whole card, which is provided by the chip) -- done.
  - [ ] Move the provenance indicators (e.g. 🤖 100%) into the "Authors" heading -- done
  - [ ] Reduce and unify vertical space and padding in node cards -- _have set a height for the details sections so all sections, when collapsed, are the same height, have reduced that height by a couple of pixels, and set all sections to be collapse by default_
  - [ ] Remove double borders within node card sections -- _done_
  - [ ] Change the font color in node cards to be a dark shade of the base card color e.g. dark blue for paragraphs (do this allowing for likelihood of at least some card colors changing) -- _done, for all text and icons (except the execution button, but can do this as well)_
  - [ ] For executable nodes, improve the appearance of execution status when the not has not yet been executed -- _not really sure how much / what kind of improvements were metnioned for this, but have added the 'Never Executed' string for now, and happy for a quick consult on the rest :)_
  - [ ] For executable nodes, simplify the execution buttons to a single "Run" button with a `<sl-dropdown>`  for (Run above, Run below etc) -- _have more or less completed the base of this, the commands need to be updated I think, and there is some display/position issues with the dropdown container being overlayed by other node cards_
  - [ ] For inline nodes, make the node card wider (currently constrained by the tooltip width?)  -- _have set min-width on the container to icrease the base width, max-width has also been set_

----------------------------

**related issue**  #2251


mike-parkin:
- [ ] For executable nodes, simplify the execution buttons to a single "Run" button with a `<sl-dropdown>`  for (Run above, Run below etc) -- _have more or less completed the base of this, the commands need to be updated I think, and there is some display/position issues with the dropdown container being overlayed by other node cards_
  
 Have resovled this issue, however to do so I had to remove/replace some css rules that I assume were set up to initiate hardware acceleration for performance reasons.... these rules also create new element stacking (ie. ignores the z-index of other elements).
 
This shouldn't cause any problems but just want to note somewhere that this has been done (specifically in `web/src/ui/animation/collapsible.ts`).

mike-parkin:
Have pushed up those minor changes we discussed earlier:

- execution details: "Never executed" is now "Not executed", and all empty details are hidden until executed.
- add text colour to AI author icons
- provenance counts: replace the black & white with the lighter node ui colours.

PR #2284, opened 2024-07-04
fix(if block): improve web component

nokome:
Our designs for `IfBlock`s look like this:

image

My first iteration towards those designs are:

Screenshot from 2024-07-05 10-44-35

Note that we do not need the editing elements such as trash can, drag and drop grips, or language select right now.

@mike-parkin Could you take this over. There are several TODOs in `web/src/nodes/if-block.ts` and `web/src/nodes/if-block-clause.ts`.

Also, it would be good to fix this z-index issue if possible:

Screenshot from 2024-07-05 10-25-14

### DIFF
--- a/examples/kernels/if-block/standalone-polyglot.json
+++ b/examples/kernels/if-block/standalone-polyglot.json
@@ -8,7 +8,51 @@
       "clauses": [
         {
           "type": "IfBlockClause",
-          "code": "false",
+          "code": {
+            "string": "false"
+          },
+          "programmingLanguage": "js",
+          "isActive": false,
+          "content": [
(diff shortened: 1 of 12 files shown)


nokome:
Thanks @mike-parkin. This is looking good, some things that would be good to tidy up before merging:

- [ ] code expression to be full width as in designs
- [ ] icons for the `if`s, `elif`s and `else` as in designs

I did some work to improve the language icons and display name and have pushed that.

image

mike-parkin:
Have implemented the requested changes

image

nokome:
Thanks Mike!

Issue #2280, opened 2024-07-03
feat(web): add styling and folding for admonitions


nokome:
During testing I noticed that `<stencila-admonition>` was missing so have stubbed one out with a bunch of `TODOs` in it related to styling:

https://github.com/stencila/stencila/blob/main/web/src/nodes/admonition.ts

Currently it looks like this:

![image](https://github.com/stencila/stencila/assets/1152336/dd4682cc-0467-4c4f-81cb-adeeb94ed612)

But should look more like all the other admonitions, "callouts", and "alert boxes" on the web e.g.

- https://v1.tailwindcss.com/components/alerts#titled
- https://flowbite.com/docs/components/alerts/
- https://docusaurus.io/docs/markdown-features/admonitions
- https://mystmd.org/guide/admonitions

mike-parkin:
@nokome thanks for the TODOs, would you like me to treat these nested paragraph nodes the same as list items, table cells captions etc...

mike-parkin:
@nokome here is the "draft" style I've chosen for the admonition

![image](https://github.com/stencila/stencila/assets/73506758/68b61c9d-7ea1-4035-85e1-fa76a8d8e371)

No rush, but when you get round to it,
Just leave a list of any requested changes in a comment here :)

nokome:
Thanks @mike-parkin 

> treat these nested paragraph nodes the same as list items, table cells captions

Yes, that seems appropriate.

> Just leave a list of any requested changes in a comment here

I feel like the borders could be a little more rounded and a little lighter maybe? But maybe get a second, or third opinion.

PR #2285, opened 2024-07-05
Feat/style admonition node

mike-parkin:
**details**

Styles the admonition nodes,

 - colours and icons set for the various types of Admonition (mapped in the `ui/icons-and-colours.ts` utility)
 - collapsible content (if initiated)
 - if no admonition title, the slot is filled with the 'admonitionType' string

Re. the comments from the issue (#2280) I did round the corners a bit more, and dropped the border colour down a notch... and I personally think it looks much better, so it was a good suggestion (no one around the office today to get a good second opinion from, but can wait till monday for that )

image

**related issue**: #2280


nokome (approved):
Thanks @mike-parkin. Looks good. I did some refactoring to reduce the amount of code (and tested it still works). Will merge now.

PR #2287, opened 2024-07-19
refactor: use inline for code output datatype nodes

mike-parkin:
**details**

Have swapped out the data type nodes form the block display to an inline, this allows them to be used in paragraphs etc without breaking the document flow.


nokome:
Thanks @mike-parkin . This is a great fix to have in place, specifically because it means the outputs of code expressions are inline.

PR #2288, opened 2024-07-19
fix(code): z-indexing

mike-parkin:
**details**

fixes the code snippet z-indexing problem when the linting tooltip crosses into another code snippet Re. the if block clauses


nokome:
Thanks @mike-parkin . Can confirm this works as intended:

image

PR #2291, opened 2024-07-21
fix(for loops): style component towards designs

nokome:
This styles component as per designs (well close to them). Currently looks like below. @mike-parkin could you pick this up and fix the following:

- [x] Iteration headers should not show when the parent `ForBlock` card is not expanded (as done for `IfBlocks`)
- [ ] Render `programmingLanguage` (as done for `IfBlockClause`)
- [x] Expand width of `code` to take up available space (as done for `IfBlockClause`)
- [x] Have a toggle-able caret for each iteration allowing content to be collapsed or expanded, defaulting to expanded
- [ ] Any other improvements you think are necessary.

image

image

### DIFF
--- a/examples/kernels/for-block/assigned-quickjs.md
+++ b/examples/kernels/for-block/assigned-quickjs.md
@@ -1,5 +1,5 @@
 ```quickjs exec
-const items = [1, 2, 3]
+var items = [1, 2, 3]
 ```
 
 ::: for item in items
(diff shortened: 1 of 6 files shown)


mike-parkin:
This is ready now, sans the programming language icons, can integrate the programming language later after the include / call blocks are done.

One extra addition I could make is to add the item / value to the iteration header, If you think its a good idea then I'll need the var, and the iterable value added to the section somehow.

image

nokome:
Thanks @mike-parkin . I've merged this, I don't think there is a need to add the iteration values, apart from anything else they maybe quite complex objects such as a database row etc.

PR #2294, opened 2024-07-22
fix(include & call blocks): improve and add web components

nokome:
This PR improves the web component for call blocks:

image

image

It also adds components for `CallBlock` and `CallArgument` nodes:

image

This PR also does some refactoring and refinements that affected `IfBlock`. So, for the record here's what that that looks like:

image

I've also done some other refactoring and tidy-ups along the way.

@mike-parkin could you pick this up and fix the following:

- [ ] Improve horizontal with and vertical alignment of `CallArgument`s
- [ ] Improve the appearance of `<stencila-ui-node-execution-message>` (see screenshot above). Now that we have code related execution messages displayed using CodeMirror, I think we should make the display of these not-code-related messages displayed as similar as possible to them (i.e. colored left border and `font-sans`.
- [ ] Any other improvements you think are necessary.

The relevant test examples are:

- http://localhost:9000/examples/nodes/include-block/include-block.md?view=dynamic
- http://localhost:9000/examples/nodes/call-block/call-block.md?view=dynamic

Note that execution of `CallBlock`s is not implemented on the backend yet so running them will not do anything.


mike-parkin:
@nokome would you like the execution messages inside a collapsible field?

nokome:
> @nokome would you like the execution messages inside a collapsible field?

Yes, that makes sense

mike-parkin:
image

Have styled the src and args like so, note, I've gone to set an arbitrary max width on all the inline codemirrors instances to avoid them pushing the other elements outside the card. and the default behaviour for codemirror is to scroll the x overflow.

PR #2296, opened 2024-07-30
Feature/add commands to suggestion blocks

mike-parkin:
**details**

- hides suggestions where status is not "Proposed" or `undefined`
- adds in the `duration` and `ended` execution details to the card
- adds command buttons to the card header
- add input for the revise

I have linked up the commands to the buttons, but only the 'reject' seems to be working... accept will return an error stating a missing `nodeId` and I'm not sure how to handle the revise yet

### DIFF
--- a/web/src/clients/commands.ts
+++ b/web/src/clients/commands.ts
@@ -20,6 +20,9 @@ export interface DocumentCommand {
     | 'execute-nodes'
     | 'interrupt-document'
     | 'interrupt-nodes'
+    | 'accept-node'
+    | 'reject-node'
+    | 'revise-node'
 
   /**
    * The type of the node that the command is being executed on.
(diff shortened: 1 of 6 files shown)


mike-parkin:
Hi @nokome , This is incomplete, but would like you to have a quick look now anyway as there are a couple of things to check.

Should hopefully check in with you in person, but will make notes here.

when clicking the `accept` button in the card, it is returning a console error about missing the nodeId when I click it? maybe this is because the suggestion node is being taken out of the DOM?

I have wired up the revise, and am passing through the instruction as `text` in the event detail, so would like to check in with you on how to get this fully wired up, currently it dispatches the event, and stays in perpetual 'pending' state on the card (I'm assuming that when the assistant has responded the webview will reload anyway).

also were the instruction messages always not showing, or is this something I have created somehow?

mike-parkin:
Have implemented the following:

- remove 'hide suggestions' toggle form instruction block
- made the 'revise' button tabbable
- enter key in text area will also trigger submission (shift+enter will use default behaviour)
- outline on input will be black
- reduce padding in revision popup

nokome (approved):

Issue #2312, opened 2024-08-29
feat(web): add a TOC for intra-doc navigation


nokome:
## Status

A new property `headings` has been added to `Article`. This property is populated, during the "compile" phase of the execution engine, with a nested `List` with `Link`s to each heading in the document.

This property is encoding to "DOM HTML" (what we send to the browser) as a `<nav slot=headings>` e.g.

```sh
cargo run -p cli compile examples/nodes/article-headings/article-headings.md --to dom --pretty
```

```html
<stencila-article id=art_FqCiGKhewxKksRZdKTWXxk depth=0 ancestors='' root>
  <nav slot=headings>
    <stencila-list id=lst_KpMnZyxCsxxh7mzhamemPY depth=1 ancestors=Article order=Ascending>
      <ol slot=items>
        <stencila-list-item id=lsi_5k3b3ptdrAFGcDgGY5v9g8 depth=2 ancestors=Article.List>
          <li slot=content>
            <stencila-paragraph id=pgh_6LHCuTvgJswcDBDGrVVsYK depth=3 ancestors=Article.List.ListItem>
              <p slot=content>
                <stencila-link id=lin_XQZU7oNJYYRzjJphfVVSv1 depth=4 ancestors=Article.List.ListItem.Paragraph>
                  <a href=#hea_LfoiH3BQLetMhigF2FE4RR>
                    <span slot=content>
                      <stencila-text id=txt_CyWbnNfT5VayZzYmZmJJ8z depth=5 ancestors=Article.List.ListItem.Paragraph.Link>Heading 1</stencila-text>
                    </span>
                  </a>
                </stencila-link>
              </p>
            </stencila-paragraph>
...
```

Currently this property is slotted in such a way that is is not displayed:

```jsx
      <!-- TODO: <stencila-ui-article-headings> component -->
      <div style="display:none">
        <slot name="headings"></slot>
      </div>
```

But, if you remove the `display:none` you can see the nested list of links to headings rendered in the browser:

![image](https://github.com/user-attachments/assets/37661e3f-78d1-4f36-a7e5-8489e3443fe8)

(Obviously we currently have `<stencila-list>` components rendering withing the headings list, which we will have to reverse).

## Inspo

There are plenty of examples of HTML TOCs around. Many are similar to Mintlify's e.g. https://mintlify.com/docs/text

![image](https://github.com/user-attachments/assets/709b58cb-71bf-40cc-b387-f35a9b3fe159)

[...]

PR #2315, opened 2024-09-05
Feature - document toc with article headings 2312

mike-parkin:
**details**

Have added a <stencila-ui-article-headings> element which renders the 'headers' slot as the docuement context in a sidebar.

 - at the article level, there is an event listener for scrolling which tells the nav element which header have been scrolled past (via a new context object).
 - the sidebar will only appear on screens larger tham ~1000px (tw's `lg` breakpoint)
 - Aside from the display and position rules, all css of the sidebar items should be set in the default.css

questions:
- colors for hover, and for the scollled states of the menu items. (currently blue, and grey respectively)
- which side shoulf the menu appear on - right seems natural, but it does interfere with the other menu. so maybe that should be moved (or integrated together maybe?).

image

image

related issue #2312

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -973,7 +973,7 @@ checksum = "1462739cb27611015575c0c11df5df7601141071f07518d56fcc1be504cbec97"
 
 [[package]]
 name = "cli"
-version = "2.0.0-beta.2"
+version = "2.0.0-beta.3"
 dependencies = [
  "app",
  "cli-utils",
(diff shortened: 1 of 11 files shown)


nokome:
Thanks Mike. I have done some documentation and renaming as I went through and reviewed this. I also changed the CSS, and modified the scroll handler so that the headings within the viewport are highlighted in a darker color:

[Screencast from 2024-09-06 11-31-57.webm](https://github.com/user-attachments/assets/691a964f-3f85-4d1e-9de4-ae150a774b43)

There is still some work to do to update the components when the headings in the document change (i.e. add some mutation observers). I can do that if you could spend some time seeing if we can achieve the Clerk style "snake".

PR #2320, opened 2024-09-12
fix(node chips): adjust vertical positioning of code chips to avoid overlaps

nokome:
Currently just stubs out a new `<stencila-ui-node-chip>` element and the use of an `IntersectionController` to detect when a chip enters or exits the viewport. Collision detection, adjustment of position, and refactoring of node cards to display and use these cards still to be done.


mike-parkin:
I've updated the elements entry callback, to create a copy of, append itself, and sort in top-down order.

It will then check itself against, the previous, and next chips in the ordered array (if they exist), this is to account for items being pushed to the array asynchronously (and not in Dom order) upon initial load.

It will only offset once, so if it offsets for the previous chip, it will skip the next.

Once an offset has been applied, we check the final position is still in the viewport, if not, we do return without adding the chip the static array.

I still feel like there's still a piece missing here, but can't quite put my finger on it where, so maybe playing around with it a bit will reveal more.

have left comments in the code as well :)

nokome:
Thanks Mike, I've merge this as it. I think there are still some improvements to be made but this is an improvement.

Issue #2326, opened 2024-10-06
fix(web): improve `DocumentMenu`


nokome:
The current `DocumentMenu` component is preliminary and needs to be refined in terms of both UI and UX.

- [x] so that there at least _something_ visible when the document is empty (or has code chunks with no outputs yet) show the Stencila icon in grey scale in the top right at all times

- [x] on hover the icon should go to full colour and the menu dropdown

- [x] rewrite the menu using Shoelace with prefix icons and dividers e.g. https://shoelace.style/components/menu-item#prefix-and-suffix

- [x] work out what should be in the menu (e.g. in addition to what is there, Run, Export) in various contexts (ie. within VSCode preview vs on the web)

The Stencila logo mark is below. I think this should be tweaked to remove the nested `<svg>` elems, the negative viewbox (IDK why its like that) and the width and height, and then added as a file to `web/src/ui/icons`. See the notes at top of `web/src/ui/icons/icon.ts`

```svg
      <svg xmlns="http://www.w3.org/2000/svg" width="60" height="60">
        <svg aria-hidden="true" viewBox="0 -23 110 110">
          <path
            d="m0 0v64h64v-64zm32 49.0667c-9.4933 0-17.1733-7.68-17.1733-17.1734 0-9.4933 7.68-17.1733 17.1733-17.1733s17.1733 7.68 17.1733 17.1733c0 9.4934-7.68 17.1734-17.1733 17.1734z"
            fill="#6f6"
          />
          <path
            d="m46.5067 49.1733c9.4845 0 17.1733-7.6887 17.1733-17.1733s-7.6888-17.1733-17.1733-17.1733c-9.4846 0-17.1734 7.6887-17.1734 17.1733s7.6888 17.1733 17.1734 17.1733z"
            fill="#2568ef"
          />
          <path
            d="m46.5066 14.8267c-2.6667 0-5.2267.64-7.36 1.7066 5.76 2.7734 9.7067 8.64 9.7067 15.4667s-3.9467 12.6933-9.7067 15.4667c2.24 1.0666 4.8 1.7066 7.36 1.7066 9.4933 0 17.1733-7.68 17.1733-17.1733s-7.7866-17.1733-17.1733-17.1733z"
            fill="#0f6860"
          />
        </svg>
      </svg>
```

PR #2332, opened 2024-10-07
Refactor/document preview menu 2326

mike-parkin:
**details**

Have refactored the document menu as per the video below

I have added in some placeholder icons and refactored all the menu into shoelace items

https://github.com/user-attachments/assets/5c337dda-a710-4668-bff6-894e5edb70ee

note the video shows a really gnarly hover state this is a firefox issue, seems to go into keyboard user mode when the page is focused so gives those high contrast focus states, but it looks fine in chrome (ie. vs code), if you like I can hack this out for now

**related issue**: #2326


nokome:
Thanks @mike-p, I've made a few minor tweaks and will merge.

PR #2342, opened 2024-10-09
refactor: apply new styles to article author provenance

### LINKED ISSUE #2339: fix(web): improve layout of document level authors and provenance
When selecting "Show authors and provenance" from the document menu you get want you want but the layout is poor - no vertical spacing between authors and the provenance is not horizontal and has no label:

image

mike-parkin:
**details**

Have applied some simple formatting style rules and labeling to the article level authors provenance slotted elements.

These should be similar to the node card displays of the same elements.

Also removed deprecated code from the `UIAuthorsProvenance` class

image

related issues: closes #2339

### DIFF
--- a/web/src/nodes/article.ts
+++ b/web/src/nodes/article.ts
@@ -1,5 +1,5 @@
 import { provide } from '@lit/context'
-import { LitElement, html } from 'lit'
+import { LitElement, css, html } from 'lit'
 import { customElement, property } from 'lit/decorators'
 
 import { withTwind } from '../twind'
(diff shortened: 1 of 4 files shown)


nokome:
Thanks @mike-parkin. Looks good but could you make the colour of the provenance pills the same hue as the background (grey), like we have in the node cards.

mike-parkin:
have added an `Artilcle` key to the ui nodes and colours mapping, and am using this to keep colour patterns consistent.

image

nokome (approved):
Awesome, thanks!

Issue #2336, opened 2024-10-08
feat(vscode extension): vscode user onboarding and initial experience 


vijay-prema:
### Summary

The vscode extension now allows sign in with Stencila for cloud features including usage of LLM credits for prompts in the document.  It mainly works on the "happy-path" but the experience is not as smooth and user-friendly as it could be.

## Issues/TODO:
- [ ] After generating the otc and before passing back to vscode, we should not show "Don't have a subscription yet?" if the user already has an active subscription/trial.
- [ ] There should be a friendly way to sign up for a subscription plan if they are logged in e.g. in the extension menu.
- [x] Saving a .smd file with spaces in the file name causes encoding of spaces to `%20`, saving a blank .smd file, and a sidecar zip with incorrect content based on incorrectly saved the blank .smd file.
- [x] After installing the extension, the user should be guided to create a login if they want to use the LLM/Router functionality, and they should be given a small number of free credits (e.g. $1 or less) without subscribing or entering CC details, such that they can at least complete the walkthroughs and try it out a handful of times.
- [x] When a user tries to sign in, then they cancel the sign in, but then try it again later, the link no longer opens.
- [x] When a user tries to sign in, and they see the sign in page and click "sign up" (as they dont yet have an account) and continue, they are force redirected to the welcome questionare and the url loses the `redirect_url` query string and so the callback to sign in to vscode never gets completed.  The workaround is to then restart vscode and sign in again, which is not obvious.
- [x] Logout (and maybe login?) usually requires the vscode to be restarted.  Could be Linux-specfic.  Perhaps just add a popup telling the user to restart vscode.
- [ ] The tutorial could be a little more clear and simple:
  - [ ] The 'Add ...' links look like web links instead of actions that add code to the current document, maybe add icons or emoji to make it more clear what it does (e.g. a pen icon).
  - [ ] Add "Learn More" button linking to docs or further examples of how to use the Instruction Block at the end.
  - [ ] Format the text better - remove bullet points and maybe break up into smaller steps or explanations
  - [ ] Add a note that you might need to log in and/or subscribe to use the LLM (or provide a local LLM or other API keys?)
- [x] Selecting the _Stencila Icon Theme_ seems to cause the icons to vanish in the file explorer.
- [x] Test out all non happy-path possibilities, creation of account from scratch vs already created etc
- [x] Error messages due to not being signed in are not friendly.  e.g.
    ```
[...]

PR #2343, opened 2024-10-09
feat(vscode ext): sign in with access token

vijay-prema:
https://github.com/stencila/stencila/issues/2336


nokome (approved):
Thanks for this. Have made some minor tweaks to the menu items and will merge.

PR #2345, opened 2024-10-10
fix(image-objects): reset all styles on svg el to avoid prose inherit

mike-parkin:
**details**

The prose typograhpy from the default theme was leaking into svg elements inside the image-object nodes, specifically mermaid.

I have set the rule `all: initial` on the svg element to override this, thus reseting the line height.

image

### DIFF
--- a/vscode/package-lock.json
+++ b/vscode/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "stencila",
-  "version": "0.0.1",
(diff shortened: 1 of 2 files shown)


nokome (approved):
Great, thanks!

nokome:
Also, love the test example 🎧 . I can see a competition evolving here for the funniest test examples.

Issue #2346, opened 2024-10-10
fix(web): tooltips for author `last-modified` are offset


nokome:
See screencast below:

[Screencast from 2024-10-10 14-53-16.webm](https://github.com/user-attachments/assets/1d14c399-6651-4cf4-a1fe-0ad7545ba75c)

PR #2351, opened 2024-10-11
fix(author): add correct positioning to last-modified tt

mike-parkin:
**details**

Luckily turned out to be quite a simple fix, 

Just added a position rule to the containing div element and seems to fix the problem.

closes: #2346


nokome (approved):
Thanks

PR #2362, opened 2024-10-15
fix(vscode): show account email if available, when account logged in

vijay-prema:
https://github.com/stencila/stencila/issues/2336

Ideally we would show the users name when logged in, but usually they have not entered it and it falls back to userId. This PR will try to find an email address to fall back to before resorting to userId.

image


nokome (approved):
LGTM, thanks!

Issue #2352, opened 2024-10-11
feat(document menu): add expand all option to the preview menu


mike-parkin:
### Summary

Add the ''Expand All Node Cards" option to the preview menu,

Upon clicking. all node cards should be expanded. 

For simplicity this should be an isolated function to expand all the cards this will be the easiest way to  avoid any conflicting states with the node chips.

This will be useful for when e2e tests are integrated so we do not have find and open the cards in before each test case.

Would recommend adding "collapse all" option as well.

nokome:
Closed

PR #2366, opened 2024-10-16
Feat/expand all node cards 2352

mike-parkin:
**details**

Add expand/collapse all node card functionality to the document via the menu.

This triggers by finding all the stencila nodes, getting the card element, and using the card's public methods for opening and closing the cards externaly.

Am marking this as draft for now as I would like some UX feedback Re. the behavior of these expanding / collapsing node cards vs the node chip state. I have set it to be relatively independent. eg. if you have the node ships on 'hide all', you can still expand via the menu. 

The performance of the 'expand all' does get quite laggy on larger documents, though this is not unexpected due to the amount of rendering required at once. 
One way to tighten this could be exploring a similar method to the nodechips checking whether or not they are within a certain offset of the vp and only expanding then, though this will require some thought.

Also noting here that I tried this with document/window event dispatches and listeners as well, the performance was pretty much the same. but have opted for this current approach to avoid having potentially hundreds of elements creating event listeners on the document.

related issue: #2352


nokome:
Thanks @mike-parkin. I quickly spun this up and tested (on a small doc). Seems fine to me. I think the node card expand/collapse can be independent of the node chip show/hide.

Some comments re the menu:

image

- seems like it is not using Inter font (?)
- there is a lot of vertical spacing; could be made more compact
- the hide/show node chips and expand/collapse node card items could both be made toggles (ie only 2 menu items for those instead of 4)

@kusalekanayake Could you give your thoughts on this?

mike-parkin:
Looks like the Inter sans font is being used (though could be a 'works on my machine' situation), at least the variable is in there.

image

kusalekanayake:
> Thanks @mike-parkin. I quickly spun this up and tested (on a small doc). Seems fine to me. I think the node card expand/collapse can be independent of the node chip show/hide.
> 
> Some comments re the menu:
> 
> image
> 
> * seems like it is not using Inter font (?)
> * there is a lot of vertical spacing; could be made more compact
> * the hide/show node chips and expand/collapse node card items could both be made toggles (ie only 2 menu items for those instead of 4)
> 
> @kusalekanayake Could you give your thoughts on this?

Hey @nokome, just getting to this now. 

I get what you mean and I think it's good to be cautious of having a growing list that's starts working against the user (like we were discussing today with the code lenses). For this menu specifically though I think we're in the clear right now with the current design. 

The 5 items we have there within `Node chips` and `Node cards` feel like they all represent actions (asides from "show on hover"). It makes me think we might get into sticky situations if we were to try group them into switches/toggles.
Say a user decides to "collapse all" the cards and sets the switch to the "collapsed" state, but then goes to expand a card or two, the switch wouldn't be able to accurately represent that state. 

Saying that, we could _maaaaybe_ look into something like grouping the opposing actions together, but I'm a little hesitant that even though we might save on vertical space, we're increasing general clutter with more lines. it's probably worth looking into again once we have a few more items in the list. 
image

nokome:
Yep fair points. Lets stick with current layout and maybe just reduce the vertical spacing a little.

mike-parkin:
have dropped the vertical spacing between the items a tad

before:

image

after:

image

nokome (approved):

PR #2367, opened 2024-10-16
Feat: setup playwright testing env.

mike-parkin:
**details**

This is the base setup at the moment. and would like to get this checked off before proceeding further.

- Have installed playwright and the corresponding argos-ci package in the project root
- Have setup the playwright-config file; added argos as a reporter.
- Have setup a couple of simple tests which are in `e2e-tests`. One of which takes an argos screenshot.
- and a new `examples/tests` directory and put some markdown files 
- Added a github workflow for playwright. This is currently in its own workflow file, and I'm thinking this should be integrated into the `test.yml` workflow. -- have not run this pipeline yet.
- Currently only using the chromium browser test, but you can uncomment the firefox and webkit in the playwright.config to use all.

To setup this first do `npm i` in the root then `npx playwright install --with-deps` (this should install the browser dependencies). 
and to run the tests is `npx playwright test`, add the `-ui` flag to get the visual.
Results will appear in the unchecked `test-results/` directory in the root, failed tests will provide screenshots.


nokome:
@mike-parkin As discussed, I have added a `e2e/examples.spec.ts` that iterates over the `.smd` files in `examples/web` and screeshots them in "plain" mode and with `?mode=test-expand-all`. If you could check for that query param and expand all node cards and all properties within node cards that would be ideal.

nokome:
Oh also, I realized that because to have humanized times e.g. "42 minutes ago" in places. I wonder if we need to make the relevant component be `?mode=test` aware. Or what might be better (not polluting prod code with test specifics) would be to [`addStyleTag`](https://playwright.dev/docs/api/class-page#page-add-style-tag) with a stylesheet that hides those elements (but..Shadow DOM 🤔 ).

mike-parkin:
@nokome  I have added a utility function `getModeParam` which takes the window instance and returns the `?mode` value or null. And have used this in the toggle chips. If  Would you like this added to the collapsible fields within the node cards as well. 

Re. the time component, I think this could be done with a css var on the relevent components eg `display: var(--timestamp-display)` as these can move through shadow doms. will look at this now.

mike-parkin:
have extended the 'test-expand-all' mode into the node card collapsible properties as well.

and gave used a css var for the above.

this is set in the default.css as `--ui-lastmod-display: block;`

and injected into the test page with 
```
const testExpandAllStyleTag = `
  :root { 
    --lastmod-display: none; 
  }
`

...... ......

      if (mode === 'test-expand-all') {
        await page.addStyleTag({
          content: testExpandAllStyleTag,
        })
      }
 ```

Here is the result in the test ui

image

nokome:
Thanks @mike-parkin . I've merged in the "expand/collapse all" PR and there is now a merge conflict on this PR. Could you fix that and then I'll merge in and setup a CI job in another PR.

mike-parkin:
@nokome conflict has been dealt with :)

PR #2392, opened 2024-11-04
docs(stencila markdown) - First two kernel help files

jduckles:
Getting started with kernel help files.


nokome:
I suggest that, at least in the short term, we repurpose this content as walkthroughs. The reason being that, we have a pathway to distributing those now, in app, rather than waiting weeks (?) for us to start self-publishing our docs.

We can open the walkthrough when the user clicks on the `?` icon on the kernel in the sidebar.

PR #2394, opened 2024-11-05
feature(kernel-r): More robust R kernel support for renv

### LINKED ISSUE #2389: feature(kernel-r) - R kernel support for renv
Playing around with R environments (📦 {renv}). 

When creating / saving a `.smd` file to a directory with a `.Rprofile` adjacent in the directory loads the profile and bootstraps`renv/activate.R`. 

It does not work up the directory tree though. So we may want to add similar logic that you have for Python there. 

Looking for the most adjacent `.Rprofile` up the tree seems reasonable. 

Allowing the user to specify one seems reasonable as well: this user option would be a forthcoming configuration decision related to #2387

nokome:
Closes #2389

@jduckles I believe that this will provide what you want for `renv` support but have not checked that yet. Can you please do that.

### DIFF
--- a/examples/workspaces/rprofiles/.Rprofile
+++ b/examples/workspaces/rprofiles/.Rprofile
@@ -1,0 +1,1 @@
+current_profile = "root-profile"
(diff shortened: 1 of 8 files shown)


jduckles:
Runs fine in base directory, created a directory `foo_bar/` and I get in that directory: 

image

Looking at the source I can't see what might be making the kernel crap out. 

Setting an invalid `R_PROFILE_USER` doesn't cause problems... 

    export R_PROFILE_USER=$HOME/test_fasdfaklsproject/; Rscript $HOME/Library/Application\ Support/io.stencila.stencila/kernels/r

PR #2398, opened 2024-11-06
refactor(markdown syntax): revise syntax for self-closing and next-block commands

nokome:
Stencila Markdown uses two shorthands for semicolon fences: "self closing" (no child blocks in content or active suggestion) and "next-block" (there is only one child block in content or active suggestion). The `docs/walkthroughs/command-syntax.smd` (see the files changed in the PR for quick access) file explains the rationale for having these shorthands.

User tests indicate that the existing syntax is confusing:

- inconsistency of having three colons at the start of the line and two angle brackets at the end e.g. `::: edit expand to 2 paragraphs >>`

- difficulty in remembering which direction angle brackets should be pointing for self-closing shorthand (`<<`) versus next-block-only shorthand (`>>`)

This PR proposes to change the syntax to:

- for self-closing fences using three colons at the end of the line e.g. `::: create code to plot data :::`

- for next-block fences using three right pointing angle brackets e.g. `::: edit expand to 40-50 words >>>`

See the changes to `docs/walkthroughs/command-syntax.smd` for more example of how this would look.

If the majority of reviewers think this is an improvement then I will extend this PR with the further changes needed to the Markdown codec, syntax highlighter, snippets, tests etc.

### DIFF
--- a/docs/walkthroughs/command-syntax.smd
+++ b/docs/walkthroughs/command-syntax.smd
@@ -41,10 +41,10 @@ Rather than having to write three new lines for a `create` command e.g.
 :::
 ```
 
-Stencila allows you to write a "self closing" opening fence by adding `<<` to the end of the line e.g.
+Stencila allows you to write a "self closing" opening fence by adding `:::` to the end of the line e.g.
 
 ```markdown
-::: create plot of height versus width <<
+::: create plot of height versus width :::
 ```
 
 ...
@@ -63,10 +63,10 @@ The cat sat on
 :::
 ```
 
-Stencila uses a "next block only" shorthand, by adding `>>` to the end of the opening fence when there is only one block in the target content or active suggestion of a command e.g.
+Stencila uses a "next block only" shorthand, by adding `>>>` to the end of the opening fence when there is only one block in the target content or active suggestion of a command e.g.
 
 ```markdown
-::: edit expand to 20-30 words >>
+::: edit expand to 20-30 words >>>
 
 The cat sat on
 ```
@@ -74,7 +74,7 @@ The cat sat on
 Here is an example of a `create` block, which has already been run and has a single suggestion:
 
 ````markdown
-::: create code to print hello world >>
+::: create code to print hello world >>>
 
 ```python exec
 print("Hello world")


yvanrichard:
I reckon it's an improvement.
What about ```::: >>> edit expand to 40-50 words```? It would feel clearer to me that it's an extension from previous.

jduckles:
After looking through this, my general question is whether or not things will be re-written by the parser when you choose one notation over the other. 

Example: If I make a block that could have had `>>>` but I use `:::` to enclose it, will the Stencila parser go and rewrite one to the other?  If so we should be cautious about which way we document and recommend, and strive to not to ever rewrite equivalent syntax the user chose. 

There are many dimensions of cognitive load going on for the user already (markdown, R/Python, HTML preview, AI/ML ....) and any time we re-write what they wrote, we increase their cognitive load wondering what just happened.

jduckles:
> I reckon it's an improvement. What about `::: >>> edit expand to 40-50 words`? It would feel clearer to me that it's an extension from previous.

FWIW a general comment, the "shorthand" is not any shorter 🤔 `:::` `:::` is the same number of characeters as `:::` `>>>`.

brettc:
1. I think the answer to "what do we gain" is less distraction, as we get less `:::` lines and less vertical whitespace. Compare:

```markdown
::: stencila command

Etiam varius odio arcu, fringilla lobortis risus convallis at.
Etiam vitae congue felis, sit amet tempus ante.
Integer eget mauris at nibh varius blandit a vitae magna.
Maecenas molestie lobortis fringilla.
Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
Aliquam ac mauris lectus.

:::

::: stencila command

Lorem ipsum dolor sit amet, consectetur adipiscing elit.
Integer sit amet nunc in mauris tristique finibus.
Vestibulum eu efficitur diam, ut maximus lacus.
Nunc sagittis nibh eu neque iaculis, quis ultrices lorem varius.
Phasellus tincidunt sapien at lobortis tristique.
Phasellus non pellentesque urna.
Nunc in pretium elit.
Integer vel nulla ullamcorper, facilisis diam sit amet, facilisis risus.

:::
```


```markdown
::: stencila command :::

Etiam varius odio arcu, fringilla lobortis risus convallis at.
Etiam vitae congue felis, sit amet tempus ante.
Integer eget mauris at nibh varius blandit a vitae magna.
Maecenas molestie lobortis fringilla.
Orci varius natoque penatibus et magnis dis parturient montes, nascetur ridiculus mus.
Aliquam ac mauris lectus.

::: stencila command :::

Lorem ipsum dolor sit amet, consectetur adipiscing elit.
Integer sit amet nunc in mauris tristique finibus.
Vestibulum eu efficitur diam, ut maximus lacus.
Nunc sagittis nibh eu neque iaculis, quis ultrices lorem varius.
Phasellus tincidunt sapien at lobortis tristique.
Phasellus non pellentesque urna.
Nunc in pretium elit.
Integer vel nulla ullamcorper, facilisis diam sit amet, facilisis risus.
```

brettc:
Looking at the above comments, it seems there are two issues here. First, the desire to simplify the syntax. The self-closing and next-block both reduce the amount of vertical white space, which is a good thing. So, even if it the same amount of characters, I think these shortcuts improve the way the document looks.

But the second issue is that the commands change the syntax when they are run. I’m still playing with this and trying to figure out what I think. Right now (with the existing syntax), it seems that Stencila accepts either `<<` or `>>` when I’m doing a `create` command. Depending on what it creates this, this could change to nothing, or `>>`. 

This does seem a lot to track. I don’t mind that *something* in the text changes when the command is executed. But I would prefer it was consistent. I need to think about this a bit more.

nokome:
Regarding re-formatting, under the current proposal, there would be two main instances of that.

When the user writes a self closing `create` command e.g.

```markdown
::: create something :::
```

after running it would be formatted to

```markdown
::: create something >>>

The suggestion
```

When the user writes a pair of colon fences because they prefer/remember that syntax

```markdown
::: create something

:::
```

after running, would be formatted to

```markdown
::: create something >>>

The suggestion
```

Possible ways to have both shortcuts available to users that want to user them *and* maintain the preference of the user that do not want to use them:

1. add a document/project/user level config option

2. add a `usedShortcuts` property to `InstructionBlock` nodes in the schema and set that on parsing and respect it when generating Markdown

I'd be very hesitant to do 2 as it would be going down the slippery slope of adding user preference to schema node types.

I think we should progress with the current proposal and, after more user testing, consider doing (1). I think that user testing with MyST where we do not have shortcuts will help. When we design the syntax for how commands will be incorporated into Quarto Markdown I suggest that we also do not use shortcuts, at least initially, and do user testing on that to see if there are complaints about the additional two lines in the document.

nokome:
This has been completed and merged to `main`.

PR #2399, opened 2024-11-07
fix(web): use `idiomorph` instead of `morphdom`

nokome:
We have recently moved to using incremental updates of the webview
in VSCode (as we do when serving from the CLI). On two occasions
(one while doing a demo) I noticed some bad updates related to the authors
of a code chunk. This was causing duplicates of the authors and severely
messing up the view. Unfortunately I am unable to replicate this right now.

I suspect (but can not confirm) that the issue may be with `morphdom`.
We are using `id`s on pretty much all elements, but it is possible that it was
messing up when it came to slots.

I thought I would try out `idiomorph` instead which uses a more robust
algorithm, which does not rely on ids being on every element. It has
been adopted by the popular Turbo framework and others and is actively
maintained.

My tests indicate it works well and I was unable to replicate the bug I
previously observed. Given the `idiomorph` should be more robust, I
suggest that we go forward with it. We have the option of reverting to
`morphdom` if we encounter any issues.


mike-parkin (approved):

mike-parkin:
My only concern was this might have an unknown effect the context mechanics wrt the EntityContext consumers
but this still all seems to be working :).

nokome:
Merged

PR #2410, opened 2024-11-12
A start on documenting of AI commands

jduckles:
Starting to document the four LLM commands with examples.


jduckles:
Linking to #2434

jduckles:
@nokome After a quick review by you I think this can be merged. Will keep adding other walkthroughs listed in #2434

jduckles:
Is any of this worth merging to add to the website @nokome

nokome:
Yes, I think so. We already have the `command-syntax.smd` walkthrough, and the `aicommands.smd` & `aicommands.myst` files we should hold off on for now I think. But the `readme.md` (aimed and walkthrough authors) and the `basic-markdown.smd` (rename to `markdown-basics.smd`?) would be good to merge now after you have done a scan to update then as needed.

PR #2420, opened 2024-11-13
Some more background and asking users to try things in walkthrough

jduckles:
Trying to develop this to be a bit more of a wholistic walkthrough with a bit more "why" at the beginning and some prompts to the user to trial commands. More to add, but this gets it started.


nokome:
Thanks for this! I've made a few tweaks to wording and ordering and will merge now. I'll also create an issue that has a checklist for the various walkthoughs that we need.

Issue #2415, opened 2024-11-13
feat(preview): Create card, re-adding revise option 


kusalekanayake:
### Which component?

Stencila VSCode extension

### What's your idea?

While adding the carousel, we removed the ability to revise a prompt in the preview UI. This is a small issue to re-add the ability to revise a prompt. It can function the way it has previously

![image](https://github.com/user-attachments/assets/8b5befea-36c8-42ad-b999-8d0af77de898)

Maybe something like this once the revise option has been selected?
![image](https://github.com/user-attachments/assets/615042ec-eea2-4247-a130-d02300b7e9e8)

nokome:
Closed in #2426

PR #2426, opened 2024-11-15
Feat/suggest block revise cmd 2415

mike-parkin:
**details**

based on this desgin below:

image

- created a new collapsible input under the suggestion header of the create block.
- The circular arrow icon to the right of the header will expand and collapse the input drawer, icon changes to chevron when the drawer is open.
- the arrow icon in the drawer will submit the revision.
- have hooked up the revise command to the lsp.
- if the input has text, it sill be submitted -- **This is not working yet**.
- have created a generic text input component for reuse in other block when needed.
- coloured the input text with the node ui `textColour`, placeholder text is the same but with .5 alpha.

_Marked as draft because I have not yet managed to get the feedback text to submit with the command, this is possibly because I used some 'legacy' code for implementation_

image

related issue: #2415


nokome:
Thanks for this @mike-parkin. I'm working on this branch now to connect up the `revise-node` command. A couple of things that I noticed also need to be done:

- it should be possible to hit enter in the revision input, instead of pressing the button
- the content of the revision input should be the `feedback` property of the current active suggestion, so that when I scroll left of right between suggestions, I can see any feedback that I previous gave

mike-parkin:
@nokome e I think I've got the 'Enter' key working properly and the inputs should display the content of the 'feedback' attribute. The only caveat to this is that if you 'dirty' the input without submitting and move to other suggestions, you'll will lose the unsubmitted feedback text.

nokome:
Thanks @mike-parkin that behaviour makes sense to me anyway.

PR #2455, opened 2024-11-28
Odt codec

mduckles:
This allows for converting to and from Open Document format (`.odt`). There are still quite a few things that are lost on conversion like the `docx` codec

### DIFF
--- a/rust/codec-odt/Cargo.toml
+++ b/rust/codec-odt/Cargo.toml
@@ -1,0 +1,8 @@
+[package]
+name = "codec-odt"
+version = "0.1.0"
+edition = "2021"
+
+[dependencies]
+codec = { path = "../codec" }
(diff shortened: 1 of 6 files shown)


nokome (changes requested):
Thanks @mduckles. I've made a few minor suggestions. In additions:

- can use please use https://www.conventionalcommits.org/en/v1.0.0/ for future commit messages - see other commits for examples

- there is an `insta` snapshot test failing. If you have already run `make -C rust fix lint test` (recommended before pushing) then the new snapshot should have been created and you can review it and accept it using `cargo insta review` (and then commit the updated snapshot file). Alternatively, you can run `car
go insta test -p codecs --test specs` to run that specific test and then do `cargo insta review`.

PR #2456, opened 2024-11-28
List formats

mduckles:
added Input and Output format rows to the table created by the `codecs` command and colors to the status 
Screenshot 2024-11-29 at 10 00 34

also added a quick message in the convert help page 
Screenshot 2024-11-29 at 10 05 02

### DIFF
--- a/rust/cli/src/convert.rs
+++ b/rust/cli/src/convert.rs
@@ -26,6 +26,8 @@ pub struct Cli {
     ///
     /// Defaults to inferring the format from the file name extension
     /// of the `input`.
+    ///
+    /// To view all supported file formats uses the codecs command.
     #[arg(long, short)]
     from: Option<String>,
 
--- a/rust/cli/src/convert.rs
+++ b/rust/cli/src/convert.rs
@@ -26,8 +26,7 @@ pub struct Cli {
     ///
     /// Defaults to inferring the format from the file name extension
     /// of the `input`.
-    ///
-    /// To view all supported file formats uses the codecs command.
+    /// See `stencila codecs list` for available formats.
     #[arg(long, short)]
     from: Option<String>,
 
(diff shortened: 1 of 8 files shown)


rust/cli/src/convert.rs
```diff
@@ -26,6 +26,8 @@ pub struct Cli {
     ///
     /// Defaults to inferring the format from the file name extension
     /// of the `input`.
+    ///
+    /// To view all supported file formats uses the codecs command.
```
nokome, line 30:
```suggestion
    /// Defaults to inferring the format from the file name extension of the `input`.
    /// See `stencila codecs list` for available formats.
```

nokome: Made some suggestions here:

- do not use a new paragraph to avoid unnecessary vertical spacing in `--help`
- use the full command and place it in backticks to make it clearer what needs to be run (we do this elsewhere too)

For consistency, this sentence should be added to all the `--from` and `--to` options in all the relevant commands `convert`, `compile`, `execute`, and `render`.

rust/codecs/src/cli.rs
```diff
@@ -38,12 +39,30 @@ struct List;
...
                 Cell::new(codec.name()).add_attribute(Attribute::Bold),
-                Cell::new(codec.status()),
+                match supports_input {
+                    true => Cell::new("Yes").fg(Color::Green),
+                    false => Cell::new("No"),
+                },
+                match supports_output {
+                    true => Cell::new("Yes").fg(Color::Green),
+                    false => Cell::new("No"),
+                },
```
nokome, line 57: Rather than `Yes/No`, I think we need a column of the formats that the codec supports and a single column indicating if it supports `to/from` those formats (I suggest those words to link to the words we are using for the `--to` and `--from` options elsewhere).

nokome (changes requested):

rust/codecs/src/cli.rs
nokome: I realized that we already had `supports_from_formats` and `supports_to_formats` so went ahead and implemented this already.

Issue #2441, opened 2024-11-24
feat(plotly): support the `plotly_mimetype` renderer


nokome:
### Which component?

Other

### What's your idea?

We need to add support for the `plotly_mimetype` renderer. Currently, without this, a browser window gets opened in a separate tab with the plot. See https://plotly.com/python/renderers/.

In the Python microkernel we need to do something like this (if `plotly` is available):

```python
import plotly.io as pio
pio.renderers.default = "browser"
```

and ensure that `fig.show` etc results in a `ImageObject` output with `mediaType: application/vnd.plotly.v1+json`.

Then, in the `ImageObject` web component we need to handle this MIME type, in a similar way that we currently handle Mermaid:

https://github.com/stencila/stencila/blob/f06315e08594e15fcf20cf4ca1bb9ff3747f40ca/web/src/nodes/image-object.ts#L55-L60

Our previous, StencilJS based, web component _might_ be useful for reference:

https://github.com/stencila/designa/blob/master/packages/components/src/components/imageDynamic/imagePlotly/imagePlotly.tsx

nokome:
Done.

PR #2459, opened 2024-11-29
feat(plots): add support for Plotly and Vega-Altiar plots 

nokome:
This add support for Plotly and Vega-Altiar plots created in Python code.

We have had this support in the past. See https://github.com/stencila/stencila/issues/2441.

@mike-parkin I have done the groundwork in the Python microkernel and added the `compilePlotly` and `compileVegaLite` methods to `web/src/nodes/image-object.ts` to `TODO`s.

See the `compileMermaid` method for ideas on how to implement these. Also previous version of this functionality _might_ be useful (but could also just be distracting and are outdated):

- https://github.com/stencila/designa/blob/master/packages/components/src/components/imageDynamic/imageVega/imageVega.tsx

- https://github.com/stencila/designa/blob/master/packages/components/src/components/imageDynamic/imagePlotly/imagePlotly.tsx


mike-parkin:
@nokome we are almost here, but there are some quicks

I have got the vega plots embedding nicely (have disabled the actions menu for now, but cna easily turn back on), would still like to be able to make them responsive.

I have the plotly rendering as an svg at the moment, becuase the plotly styles to the full canvas are applied dynamically on the document body, so they do not filter down through the shadow doms.

mike-parkin:
@nokome I have managed to get the interactive plotly figures rendering properly the image-object content, it was a little hacky but it seems to work quite nicely.

image


I have set the x-overflow on the containers to auto, so they will scroll if the window gets to small, It will be a little difficult to get them to render fully responsive t oviewport in this mode, and we still want the users to be able put in fixed widths and heights as in the screen shot above.

Currently both libraries will to default to the client dimensions of the container element at the time of render. One solution could be to check the `spec` objects for width & height parameters, and if they are not there we set an event to redraw the figure upon resize.

Issue #2453, opened 2024-11-27
bug(vscode): jumpy linked scrolling


nokome:
### What happened?

Linked scrolling is very jumpy for certain documents. I suspect it is mostly a problem for documents with nested nodes. For example, scrolling on `examples/web/multi/flat.smd` is OK, but on `examples/web/multi/nested.smd` is "jumpy".

The functionality is implemented in `vscode/src/scroll-syncer.ts` but relies on `nodeIdsForLines` and `linesForNodeIds` functions provided by the LSP to map line numbers to node ids and verse versa and `web/src/clients/webview.ts` which users observers to send messages for the editor to be able to follow the sroll position of the preview.

I suspect that what is happening is that for nested nodes, as you scroll down the `nodeIdsForLines` returns the progressively nested children and scrolls the preview to that position. But as you continue to scroll, you exit the nested child and so the `nodeIdsForLines` returns its parent and so the preview scrolls back up to the top of that parent, and so on. I think this is the main issue for the editor -> preview sync.

I have just added some settings to hopefully make testing this easier. Some of the problems could also be related to bi-directional sync of scrolling position, so I suggest turning one direction off at a time, testing and fixing it, and then testing and fixing them together.

![image](https://github.com/user-attachments/assets/6671fcec-33be-441b-b51a-24c115587ec8)

Note that currently, `editorFollowsPreview` only happens if the preview has focus. It did this to avoid conflicts with editor cursor position but that might not be necessary and I think best to avoid.

### Which product?

Stencila VSCode extension

PR #2464, opened 2024-12-01
Fix/vscode ui bugs

### LINKED ISSUE #2463: bug(suggestion-block): issues with suggestion overflow
### What happened?

I have picked up a couple of bugs with the suggestion blocks that occur when there is more that one suggestion

- when second suggestion is active in the preview, the suggestion block element is block the click events of the create card's node marker, I believe this is because the 
- The hidden overflow for the suggestion slot is still adding vertical scroll to the window, as shown in image. Unfortunately setting the `overflow-x: hidden` on the slotted element will not help here.

image

 
 

### Which product?

Stencila VSCode extension

### Which version?

0.0.18

### Log output

_No response_

mike-parkin:
**details**

Suggestions
- set overflow hidden on the parent of transforming carousel div to stop overflowed suggestions creating x scrolling space
- disable pointer events on carousel div to stop the left overflow blocking click events on the create node marker

Scroll link behaviour
- disabled the preveiw scroll linking on nested elements, this at least stops the jumpyness, but if the parent node is long then it will not track the scrolling within that element.

closes #2463 

related #2453

### DIFF
--- a/web/src/clients/webview.ts
+++ b/web/src/clients/webview.ts
@@ -365,8 +365,7 @@ export class WebViewClient {
   /**
    * Handle a received `ScrollSyncMessage` message
    */
-  private handleScrollSyncMessage(data: ScrollSyncMessage) {
-    const { startId, cursorId } = data
+  private handleScrollSyncMessage({ startId, cursorId }: ScrollSyncMessage) {
     // Prioritize cursor position if it exists
     if (cursorId) {
       const cursorElement = document.getElementById(cursorId)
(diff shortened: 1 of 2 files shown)


nokome:
> disabled the preveiw scroll linking on nested elements, this at least stops the jumpyness, but if the parent node is long then it will not track the scrolling within that element.

This is an improvement, but I think we should be able to do better. We should be able to take the mid-point line in the editor, work out which node that is in, get it's node id and the relative position between 0 and 1 and scroll to the same relative position of the bounding rect of that node id. I'll have a crack at that - it should be simpler and closer to user expectations of how scrolling should be linked.

PR #2466, opened 2024-12-02
fix(pandoc codec): add encoding/decoding of more node types

nokome:
- adds `examples/conversion` tests for Pandoc based codecs (and renames `example/nodes` to `examples/conversion`).
- implements encoding/decoding for more node types by Pandoc codec


mduckles:
For implementing the round trip for `IncludeBlock` `execution_mode` is converted from and to a string should this be abstracted to methods on `ExecutionMode` ?  https://github.com/stencila/stencila/blob/3bfe279859735ce45bc34320e364d7908bcd3882/rust/codec-pandoc/src/blocks.rs#L580-L586 https://github.com/stencila/stencila/blob/3bfe279859735ce45bc34320e364d7908bcd3882/rust/codec-pandoc/src/blocks.rs#L779-L785

nokome:
Actually, `ExecutionMode` already derives `strum::Display` and `strum::EnumString` so the to/from string methods are already there. And `strum(ascii_case_insensitive)` is used so it maybe that you can simplify to

```rs
let mode = mode.to_string().to_lowercase();
```

and

```rs
let execution_mode = ExecutionMode::from_str(execution_mode).ok(); 
```

Or something like that...

mduckles:
not sure if they should, myst and qmd don't support the round trip for abstracts at the moment.

PR #2473, opened 2024-12-08
fix(jats codec): further implementations

nokome:
This PR adds more implementations to the JATS codec. Suggested implementation order:

- [x] `List`
- [x] `MathBlock`
- [x] `CodeBlock`
- [x] `Figure`
- [ ] `Table`
- [x] `Claim`
- [x] `CodeChunk`
- [ ] `ForBlock`
- [ ] `IfBlock`
- [ ] `IncludeBlock`
- [ ] `CallBlock`
- [ ] `RawBlock`
- [ ] `StyledBlock`

Checklist for implementing each:

- Check that the `xxx.jats.xml` file for the example looks mostly correct (i.e. the correct XML tag name and attribute names)
- Remove any JATS skips in `config.yaml` for the example
- Implement `decode_xxx` in `rust/codec-jats/src/decode/body.rs`
- Run `INCLUDE_EXAMPLES=xxx make -C examples conversion` until passes (add `skip-props` in `config.yaml` if necessary)
- Remove node type from skipped types in `rust/codecs/tests/proptests.rs` and run `cargo test -p codecs --test proptests --features proptest-high jats` (add skips for properties if necessary).

See the commit 27cffa00c19f80f26ac1d546c419dfbd27208c87 for `List` as an example of the above.


mduckles:
currently `Figure`: `caption`,`label` and `label_automatically` aren't being encoded to `jats.xml`

nokome:
Of the remaining types to implement,  ForBlock, IfBlock, IncludeBlock, CallBlock, RawBlock and StyledBlock do not have direct analogues in JATS. One option is to encode them as special namespaced elements e.g. `<stencila:for-block>` but that would only be useful if JATS is a commonly used authoring format, which it is not. Instead, I think we use a "render" approach for these node types i.e. only encode `ForBlock.iterations`, `IncludeBlock.content` etc.

rust/codec-jats/src/decode/body.rs
```diff
@@ -130,19 +136,194 @@ fn decode_sec(path: &str, node: &Node, losses: &mut Losses, depth: u8) -> Block
...
+    if let Some(automatically) = node
+        .attribute("label-automatically")
+        .map(|string| match string {
+            "true" => Some(true),
+            "false" => Some(false),
+            _ => None,
+        })
+    {
+        label_automatically = automatically;
+    }
```
nokome, line 199: This should be able to be simplified to:

```suggestion
    let label_automatically = node
        .attribute("label-automatically")
        .map(|string| match string {
            "true" => Some(true),
            "false" => Some(false),
            _ => None,
        });
```

or even

```suggestion
    let label_automatically = node
        .attribute("label-automatically")
        .and_then(|string| string.parse().ok());
```

(neither tested)

nokome:
I have added some example articles from biorXiv and eLife for testing decoding articles "from the wild" - mainly for testing of `<front>` and `<back>` matter. Run snapshot tests for these using

```
cargo insta test -p codec-jats --test examples
cargo insta review
```

Looking a the example `.xml` files and the `.decode.losses` snapshots is a good way to figure out what/how to decode things.

When we have these ones in a decent state, we'll add more test examples, including from other sources.

mduckles:
Table.notes aren't currently being encoded should they?

nokome:
Thanks for the work on `<table-wrap>` @mduckles. I have done some [minor refactoring](https://github.com/stencila/stencila/pull/2473/commits/58be37f01388e4be77ef4ff551f49ae5ffe818f4) to simplify (and remove an unnecessary `extend_path`)  - please take a look for future reference.

Can you get in the habit of running the relevant tests and committing any updated files:

```sh
make -C examples conversion
cargo insta test -p codec-jats
cargo insta review
```

With the last command you get to accept, reject etc updates to the snapshots e.g.

image

I accept them (and then commit the change) if the update looks right given the changes I've just made to the code.

Currently, these snapshots indicate a couple of issues:

1. the `rust/codec-jats/tests/examples/biorxiv-573522.xml` example has a `<graphic>` in each `table-wrap` so we end up with no `TableRow`s - nothing we can do with that at this stage
2. for the eLife example the screenshot above indicates we are loosing attributes like "align", "valign", and "rowspan"
3. for the eLife example the screenshot above (and the related JSON and JATS snapshots) and the cell content ( e.g. `italic`, text() paths) - this is because we are using `decode_blocks(..)` on the cell children but they are inlines, not blocks

For (2) if think we should finally add `horizontalAlignment` and `verticalAlignment` properties to `CodeCell` (they already have `rowSpan` and `columnSpan`). I will do that. Once I have, can you please handle the decoding of those in `decode_table_cell`.

For (3) I think you can fix this in `decode_table_cell` by calling `decode_blocks` (as is currently done) and if that result is empty, calling `decode_inlines` and wrapping the result. Something like:

```rust
let mut content = decode_blocks(path, node.children(), losses, depth);
if content.is_empty() {
  content = vec![p(decode_inlines(path, node.children(), losses))];
}
```

nokome:
Added handling of table notes in f123c71dbe5800d1918bd92bb489a9d359f90fbb. Better handle of footnotes, including "table footnotes" needed in the future.

nokome:
Added horizontal and vertical alignment properties to `TableCell` in fd9073086db0cd24db530efd4ccf87b10507ee0c

rust/codec-jats/src/decode/front.rs
```diff
@@ -1,16 +1,331 @@
...
+        } else if tag == "month" {
+            month = child.text().unwrap_or_default()
+        } else if tag == "year" {
+            year = child.text().unwrap_or_default()
+        }
+    }
+    article.date_published = Some(Date {
+        value: [year, "-", month, "-", day].concat(),
+        ..Default::default()
+    });
```
nokome: This would have given invalid date strings if elements were missing or empty so I fixed it in 1250e3616239bfe6360eec012c3395486e49f6eb

nokome:
@mduckles Thanks for the updates. I've left some comments. After you have addressed those could you tackle the `<ref-list>` in the `<back>`. For Typescript implementation of that see: 

https://github.com/stencila/encoda/blob/7dd7b143d0edcafa67cab96bf21dc3c077613fcc/src/codecs/jats/index.ts#L1192

mduckles:
> @mduckles Thanks for the updates. I've left some comments. After you have addressed those could you tackle the `<ref-list>` in the `<back>`. For Typescript implementation of that see:
> 
> https://github.com/stencila/encoda/blob/7dd7b143d0edcafa67cab96bf21dc3c077613fcc/src/codecs/jats/index.ts#L1192

rust/codec-jats/.vscode/launch.json
nokome, line 1: This seems to have accidentally been committed. Probably should be removed

nokome:
Thanks for all this @mduckles. I have rebased this branch on to `main`, made some fixes to get tests and linting passing and have merged.

PR #2475, opened 2024-12-10
feat(web): model chat interface

nokome:
Adds a "Chat with Model" interface that will initially be integrated with the VSCode extension but in the future is likely to be used in other integrations (e.g. Google Docs, Curvenote).

- for when the user wants to have a conversation with a model directly rather than using a specific prompt

- content generated by the model may includes multiple node types e.g. `Paragraph`, `CodeChunk`, `MathBlock` which the user can:
  - ask the model to refine by adding another message to the chat conversation
  - edit themselves
  - preview (in the case of math Mermaid diagrams etc)
  - run (in the case of executable `CodeChunk`s)
  - insert into the document

- similar to ChatGPT interface but with key differences being:
   - user can select which model to user
   - document is automatically, but optionally, used as context
   - workspace files can be included in context
   - generated content can be edited directly and inserted into the document

In essence the implementation approach is to:

- create a "model chat" system prompt that differs from the existing builtin prompts that we have in that it:
  - provides instructions for how to generate several different types of content (e.g. `CodeChunk`, `MathBlock`)
  - does not discourage the model from generating explanations or comments
- add the current document as context to the system prompt
- add any files that the user specifies to the system prompt
- create a `ModelTask` with an `InstructionMessage` for the system prompt and each of the existing user/model messages in the chat
- send the `ModelTask` to the model
- parse the generated content as Markdown into an `Article` and add that to the chat interface


vscode/package.json
```diff
@@ -174,6 +174,13 @@
         "shortTitle": "Preview doc",
         "icon": "$(preview)"
       },
+      {
+        "command": "stencila.open-chat-assistant",
+        "category": "Stencila",
+        "title": "Open chat assistant",
+        "shortTitle": "Chat assistant",
```
nokome, line 181: We are avoiding anthropomorphism and trying to take a more sober, matter-of-fact, approach to LLM use so avoiding "assistant" and "agent" at present:

```suggestion
        "command": "stencila.open-model-chat",
        "category": "Stencila",
        "title": "Open model chat",
        "shortTitle": "Model chat",
```

vscode/package.json
```diff
@@ -600,6 +607,11 @@
           "command": "stencila.invoke.export-doc",
           "when": "editorLangId =~ /^(md|smd|myst|qmd)$/",
           "group": "000-Stencila@3"
+        },
+        {
+          "command": "stencila.open-chat-assistant",
```
nokome, line 612:
```suggestion
          "command": "stencila.open-model-chat",
```

vscode/src/commands.ts
```diff
@@ -242,4 +242,17 @@ export function registerDocumentCommands(context: vscode.ExtensionContext) {
...
+  context.subscriptions.push(
+    vscode.commands.registerCommand(
+      "stencila.open-chat-assistant",
+      async (docUri) => {
+        const editor = vscode.window.activeTextEditor;
+        if (!editor) {
+          vscode.window.showErrorMessage("No active editor");
+          return;
+        }
+        await createChatAssistentPanel(context, editor);
```
nokome, line 254:
```suggestion
      "stencila.open-model-chat",
      async (docUri) => {
        const editor = vscode.window.activeTextEditor;
        if (!editor) {
          vscode.window.showErrorMessage("No active editor");
          return;
        }
        await openModelChatPanel(context, editor);
```

vscode/src/webviews.ts
```diff
@@ -286,3 +286,76 @@ export function closeDocumentViewPanels() {
...
+export async function createChatAssistentPanel(
+  context: vscode.ExtensionContext,
+  editor: vscode.TextEditor,
+) {
+  const webDist = vscode.Uri.joinPath(context.extensionUri, "out", "web");
+
+  const panel = vscode.window.createWebviewPanel(
+    "chat-assistant-view",
+    "Chat Assistant",
+    vscode.ViewColumn.Beside,
```
nokome, line 299:
```suggestion
export async function createChatAssistentPanel(
  context: vscode.ExtensionContext,
  editor: vscode.TextEditor,
) {
  const webDist = vscode.Uri.joinPath(context.extensionUri, "out", "web");

  const panel = vscode.window.createWebviewPanel(
    "model-chat-view",
    "Model Chat",
    vscode.ViewColumn.Beside,
```

web/package.json
```diff
@@ -38,7 +38,8 @@
       "source": [
         "src/themes/default.css",
         "src/themes/stencila.css",
-        "src/views/vscode.ts"
+        "src/views/vscode.ts",
+        "src/views/vscode-chat.ts"
```
nokome, line 42: Is it best to have to separate files? (I suspect it is but just checking). If so should rename to `vscode-preview.ts` and `vscode-model-chat.ts`

web/src/clients/chat-assistant.ts
```diff
@@ -0,0 +1,64 @@
+type ReceivedMessage = AssistantResponseMessage
```
nokome, line 1: Rename file and `interface`s etc to reflect "model chat" rather than "chat assistant"

nokome (commented):
Thanks for getting this started @mike-parkin. I've left some comments mainly around naming.

I need to spend some time thinking about how best to represent the chat on the server side and do the comms back and forth to the client.

mike-parkin:
Thanks @nokome , I like modelchat better :)

web/package.json
nokome: Fair enough. I don't think there is a strong need to DRY in the near future. I generally use the [Rule of Three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming)).

PR #2477, opened 2024-12-19
feat(vscode): Eventing 

jduckles:
Eventing integration into VSCode is active here.

@nokome Should I abstract the public API key out to some configuration file?

Example PostHog event for a file export: 

image


nokome:
> Should I abstract the public API key out to some configuration file?

I don't think that is necessary, it is not really something that needs to be configurable - there is only one key that is appropriate to use - and making it configurable will only increase brittleness.

vscode/src/commands.ts
```diff
@@ -2,6 +2,16 @@ import * as vscode from "vscode";
...
+import { posthog } from 'posthog-js'
+
+
+const isTelemetryEnabled = vscode.env.isTelemetryEnabled;
+
+
+if (isTelemetryEnabled) {
+  posthog.init('LeXA_J7NbIow0-mEejPwazN7WvZCj-mFKSvLL5oM4w0', { api_host: 'https://events.stencila.cloud' })
+}
+
```
nokome, line 14: I suggest moving this to a new module `events.ts` and defining an `event` function that is conditional on `vscode.env.isTelemetryEnabled`.

```ts
export const event = isTelemetryEnabled
  ? (name: string, data?: any) => posthog.capture(name, data)
  : (name: string, data?: any) => {};
```

This will reduce the amount of boilerplate code at call sites:

```ts
 import { event } from "./events.ts"

event("vscode_preview");
```

nokome:
@jduckles I haven't merge this yet because it seemed incomplete - it only sends two events. I think it is important to add events for signin/signout of Stencila cloud, creating files of different formats, adding provider api keys etc. Maybe I should go ahead and add those I think appropriate and get you to review?

jduckles:
That works, I have listed a bunch of possibilities in the [Google Doc](https://docs.google.com/document/d/1yNApJW-fZGsqjxCPP-kC5Xqlve49Q04RPGzE19UmL0M/edit?tab=t.0), we should have the conversation about which are priorities and how we want to handle the event name namespace. I'm happy to review and test each one if you know right where to pop the `event()` calls.

nokome:
Argh, sorry, I forgot about that doc!

PR #2482, opened 2025-01-06
Fix/bump deps

mike-parkin:
have migrated eslint to the latest version and converted all the config files to the new flat config style

bumps the `parcel` and `@parcel/transformer-*` packages to latest

may need to remove the .parcel-cache and node_modules for the parcel updates to not throw errors


nokome:
Thanks @mike-parkin . I had to make some commits to get CI to pass (due to new eslint config) but working now, so will merge

PR #2489, opened 2025-01-07
feat(lexical codec): add codec for Lexical and Koenig JSON

nokome:
Work in progress.


rust/codec-lexical/tests/snapshots/examples__ghost.decode.losses.snap
```diff
@@ -0,0 +1,10 @@
+---
+source: rust/codec-lexical/tests/examples.rs
+expression: info.losses
+---
+CodeBlockNode.caption: 1
+HtmlNode.visibility: 1
+Link: 1
+List: 2
```
nokome, line 8: These need to be handled in `blocks_from_lexical`

rust/codec-lexical/tests/snapshots/examples__ghost.decode.losses.snap
```diff
@@ -0,0 +1,10 @@
...
+Unknown (aside): 1
+Unknown (image): 1
```
nokome, line 10: Need to decide how we will handle these. For `image`, `audio`, `video` and `file` Ghost Koenig cards I am leaning towards adding `ImageObject`, `AudioObject` etc to Stencila's `Block` enum, rather than wrapping in a paragraph as we currently do.

rust/codec-lexical/tests/examples/ghost.koenig
```diff
@@ -0,0 +1,470 @@
...
+          },
+          {
+            "detail": 0,
+            "format": 128,
+            "mode": "normal",
+            "style": "",
+            "text": "highlight",
+            "type": "extended-text",
+            "version": 1
+          },
```
nokome, line 185: We are not currently handling `format: 128` because do not have inline `Highlight` yet in Stencila. I suggest we add that and then update the `TextFormat` bitfield and handling.

rust/codec-lexical/tests/examples/ghost.koenig
```diff
@@ -459,6 +459,101 @@
...
+              {
+                "detail": 0,
+                "format": 0,
+                "mode": "normal",
+                "style": "",
+                "text": "https://ghost.org/help/cards/",
+                "type": "extended-text",
+                "version": 1
+              }
+            ],
```
nokome, line 485: Let's remove this TODO paragraph

jduckles: Done.

rust/codec-lexical/tests/examples/ghost.koenig
nokome: Done in [96add3e](https://github.com/stencila/stencila/pull/2489/commits/96add3e28c7c2a8511d4f25ef84783d3e2ba49f0)

schema/Annotation.yaml
nokome, line 1: @jduckles I have added this inline node type based on our discussion. It is a `Mark` in that it spans inline `content` but is similar to a `Note` (also an inline node type, used for foot nodes, side notes etc) in that it has an `annotation` property that is an array of block nodes. If necessary, more properties could be added in the future (e.g. "tags") but I have kept it simple for now.

jduckles: In your notes in the yaml I'm confused by "usually a paragraph", reading that I think your usual usage would be to only wrap the entirety of a <p> </p> equivalent. Does it work equally well to wrap some sub-string of a paragraph like I'm demonstrating below? If so, I suggest you reword "usually a paragraph".

It is an edge case but can partially overlapping annotations be supported (I'm guessing not in the current notation style)? Like:

```markdown
This is some text I'd like to add (start annotation 1)=={an annotation I would like to capture, that 
also has a (start annotation 2) =={nested key point }==(end annotation1) I'd also like to annotate.}==(end annotation2)
```

This does come up in social science "qualitative coding" more than you might think.

nokome: > Does it work equally well to wrap some sub-string of a paragraph like I'm demonstrating below?

Yep, that'll work, fine. That is lazy copy-pasta taken from the defn for `Note`.

> It is an edge case but can partially overlapping annotations be supported (I'm guessing not in the current notation style)?

No that can't be supported. It is not a limitation of a the Markdown notation but of the tree structure of the document schema. There are ways around that but they would require a _major_ re-architecturing.

Also, right now we are now we have not added Markdown support for `Annotation`s with `annotation` properties like your examples. Only highlights, with no annotation text are supported. Actually come to think of it, the Markdown "decoding" of those isn't even supported.

rust/codec-lexical/src/inlines.rs
```diff
@@ -0,0 +1,202 @@
...
+            rel: link.rel,
+            target: link.url.clone(),
+            content: vec![Inline::Text(Text::from(link.url))],
+            ..Default::default()
+        })
+    }
+}
+
+fn link_to_lexical(link: &Link) -> lexical::InlineNode {
+    if let Inline::Text(Text { value, .. }) = link.content.first().unwrap() {
```
nokome, line 188: Avoid `.unwrap()` like the plauge - this would cause a `panic` if `link.content` is empty!

```suggestion
    if let Some(Inline::Text(Text { value, .. })) = link.content.first() {
```

rust/codec-lexical/src/inlines.rs
```diff
@@ -0,0 +1,202 @@
...
+        lexical::InlineNode::Link(lexical::LinkNode {
+            url: value.to_string(),
```
nokome, line 190: I don't understand why the `link.content` is going into the `url`, shouldn't the content being converted to `children` like we do eslswhere?

```rust
let children = inlines_to_lexical(&link.content, context);
```

rust/codec-lexical/src/lexical.rs
```diff
@@ -0,0 +1,301 @@
...
+    pub text: String,
+}
+
+#[skip_serializing_none]
+#[derive(Default,Serialize, Deserialize)]
+pub(super) struct LinkNode {
+    pub r#type: MustBe!("link"),
+
+    pub format: String,
+
```
nokome, line 241: This seems to be missing `children`. This is an example from `layground.lexical`:

```json
              {
                "children": [
                  {
                    "detail": 0,
                    "format": 0,
                    "mode": "normal",
                    "style": "",
                    "text": "GitHub repository",
                    "type": "text",
                    "version": 1
                  }
                ],
                "direction": "ltr",
                "format": "",
                "indent": 0,
                "type": "link",
                "version": 1,
                "rel": null,
                "target": null,
                "title": null,
                "url": "https://github.com/facebook/lexical"
              },
```

rust/codec-lexical/src/blocks.rs
```diff
@@ -0,0 +1,409 @@
...
+        ..Default::default()
+    })
+}
+
+fn list_to_lexical(list: &List, context: &mut LexicalEncodeContext) -> lexical::BlockNode {
+    let children = list
+        .items
+        .clone()
+        .into_iter()
+        .map(|item| list_item_to_lexical(item.content.first().unwrap(), context))
```
nokome: Don't use `unwrap()` - it would cause a panic if content happens to be empty. Also, `clone()` here should not be necessary if  `list_item_to_lexical` was to take `&ListItem`.

rust/codec-lexical/src/blocks.rs
```diff
@@ -0,0 +1,409 @@
...
+        lexical::ListItemNode {
+            children: inlines_to_lexical(content, context),
+            ..Default::default()
+        }
+    } else {
+        lexical::ListItemNode {
+            ..Default::default()
+        }
+    }
+}
```
nokome: I suggest changing the above two functions to this:

```rust
fn list_to_lexical(list: &List, context: &mut LexicalEncodeContext) -> lexical::BlockNode {
    let children = list
        .items
        .iter()
        .map(|item| list_item_to_lexical(item, context))
        .collect();

    lexical::BlockNode::List(lexical::ListNode {
        children,
        ..Default::default()
    })
}

fn list_item_to_lexical(
    list_item: &ListItem,
    context: &mut LexicalEncodeContext,
) -> lexical::ListItemNode {
    let children = inlines_to_lexical(&blocks_to_inlines(list_item.content.clone()), context);

    lexical::ListItemNode {
        children,
        ..Default::default()
    }
}
```

nokome: Actually, since I already wrote it I tested it and committed, so done already.

PR #2492, opened 2025-01-09
Tests/expand web examples for e2e

mike-parkin:
**details** 

Adds some new test examples for if-block, for-block and paragraph authorship for argos screenshot testing

Have also altered some of the elements which consume context, to create the consumer object in the `updated` lifecycle hook, this allows for any provider elements to be loaded and initialised as the initial server rendering of the document was causing some child element to finish their lifecycle before the providers had initialised.


mike-parkin:
@nokome this is rebased to main, and looks good to merge :)

PR #2493, opened 2025-01-09
Publish to Ghost

timClicks:
This PR enables users to publish Stencila docs to Ghost.

Features:

- Stores secrets within the OS keyring via Stencila's secrets,

Limitations: 

- Uses the text codec to create a rough version of the doc in Ghost. Intended to be updated with `codec-lexical` when it's available.
- Lack of testing (am unfamiliar with some of the mechanics)
- Only files are supported currently, rather than directories

### DIFF
--- a/rust/publish-ghost/src/lib.rs
+++ b/rust/publish-ghost/src/lib.rs
@@ -130,7 +130,7 @@ pub struct Cli {
     /// The Ghost domain
     /// 
     /// This is the domain name of your Ghost instance, with an optional port. 
-    #[arg(long, env = "STENCILA_GHOST_DOMAIN", value_parser = parse_host)]
+    #[arg(long, env = "STENCILA_GHOST_DOMAIN", value_parser = parse_host, default_value = "ghost.io")]
     ghost: Host,
 
     /// The Ghost Admin API key
--- a/rust/publish-ghost/src/lib.rs
+++ b/rust/publish-ghost/src/lib.rs
@@ -183,16 +183,19 @@ impl Cli {
         // Send content to Ghost
         let token = generate_jwt(&self.key).context("generating JWT")?;
 
-        let url = if self.post {
-            format!("https://{}/ghost/api/admin/posts/", self.ghost)
+        // "root key" is the terminology use by Ghost/Lexical
+        let root_key = if self.post {
+            "posts"
         } else if self.page {
-            format!("https://{}/ghost/api/admin/pages/", self.ghost)
+            "pages"
         } else {
             unreachable!();
         };
 
+        let url = format!("https://{}/ghost/api/admin/{}/", self.ghost, root_key);
+
         let payload = serde_json::json!({
-            "posts" : [
+            root_key : [
                 json!({
                     "title": "WIP: Stencila",
                     "html": content,
--- a/rust/publish-ghost/src/lib.rs
+++ b/rust/publish-ghost/src/lib.rs
@@ -1,9 +1,123 @@
 use std::path::PathBuf;
 
+use codec::{Codec, EncodeOptions};
+// use codec_html::HtmlCodec;
+use codec_text::TextCodec;
 use common::{
     clap::{self, Parser},
-    eyre::Result,
+    eyre::{bail, eyre, Context, Result}, 
+    reqwest::Client,
+    serde::{Deserialize, Serialize},
+    serde_json::{self, json},
+    tracing,
 };
+use document::{CommandWait, Document};
+
+use url;
+use jsonwebtoken as jwt;
+
+const SECRET_NAME: &str = "GHOST_ADMIN_API_KEY";
+
+// use eyre::{eyre, Result};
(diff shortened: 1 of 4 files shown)


rust/publish-ghost/src/lib.rs
```diff
@@ -15,12 +128,18 @@ pub struct Cli {
     path: PathBuf,
 
     /// The Ghost domain
-    #[arg(long, env = "STENCILA_GHOST_DOMAIN")]
-    domain: bool,
+    /// 
+    /// This is the domain name of your Ghost instance, with an optional port.
```
jduckles, line 132: I would explicitly state without `https://` as I included this at first.

jduckles (commented):
Works with `--post` with `--page` I get: 

```
 HTTP 400: "Request not understood error, cannot save page.":
      {"errors":[{"message":"Request not understood error, cannot save page.","context":"No root key ('pages') provided.","type":"BadRequestError","details":null,"property":null,"help":null,"code":null,"id":"5ad2e220-cebf-11ef-a69b-75bdc16862c1","ghostErrorCode":null}]}
```

rust/publish-ghost/src/lib.rs
```diff
@@ -15,12 +128,18 @@ pub struct Cli {
...
+    ghost: Host,
 
     /// The Ghost Admin API key
-    #[arg(long, env = "STENCILA_GHOST_KEY")]
-    key: bool,
+    /// 
+    /// To create one, create a new Custom Integration under
+    /// the Integrations screen in Ghost Admin. Use the Admin API Key,
+    /// rather than the Content API Key.
+    #[arg(long, env = "STENCILA_GHOST_KEY", value_parser = parse_key)]
```
nokome, line 141: Mention in docs that can set the secret to avoid entering it each time:

```sh
stencila secrets set GHOST_ADMIN_API_KEY
```

rust/publish-ghost/src/lib.rs
```diff
@@ -1,9 +1,122 @@
...
+fn parse_key(arg: &str) -> Result<String> {
+    // Use the key provided on the command-line
+    if !arg.is_empty() {
+        return validate_key(arg);
+    }
+
+    // If not, check if it's provided as an environment variable
+    if let Ok(env_key) = std::env::var("STENCILA_GHOST_KEY") {
+        return validate_key(&env_key);
+    }
```
nokome, line 110: I thought that using `#[arg(..., env=)]` that we get this "for free" - i.e. CLAP checks for the env var. Or maybe it is not done before `parse_arg`

rust/publish-ghost/src/lib.rs
nokome: Also should have better consistency between the secret name and the env var name: maybe the secret name should be the env var name minus the `STENCILA_` prefix

rust/publish-ghost/src/lib.rs
```diff
@@ -33,6 +152,69 @@ pub struct Cli {
...
+        let response = Client::new()
+            .post(url)
+            .header("Authorization", format!("Ghost {token}"))
+            .query(&[("source", "html")])
+            .json(&payload)
+            .send()
+            .await?;
+    
+        if response.status().is_success() {
+            Ok(())
```
nokome, line 213: We will need to get the Ghost id of the page/post so we can add it to the metadata of the document.

nokome (commented):
Thanks @timClicks . Looks good. Like how you made use of the secrets! Some comments attached.

rust/publish-ghost/src/lib.rs
timClicks: I actually wasn't sure when I wrote it, so decided to include it so that the order is correct. I was worried that the env var would only be accessed by clap if `parse_arg` failed to find anything.

Keeping it in the attribute means that the documentation is rendered correctly.

nokome: Good thinking.

jduckles:
Also I've tested `--post` against my own self-hosted version of Ghost and it works as expected.

rust/publish-ghost/src/lib.rs
```diff
@@ -130,7 +130,7 @@ pub struct Cli {
     /// The Ghost domain
     /// 
     /// This is the domain name of your Ghost instance, with an optional port. 
-    #[arg(long, env = "STENCILA_GHOST_DOMAIN", value_parser = parse_host)]
+    #[arg(long, env = "STENCILA_GHOST_DOMAIN", value_parser = parse_host, default_value = "ghost.io")]
```
jduckles, line 133: A default of `ghost.io` won't work for anyone, it will always need a subdomain specified. So this isn't a sensible default value to include.

nokome: Right, my bad.

nokome:
Thanks for the changes @timClicks. I'll merge this now so we can combine it with the new Lexical codec and create a new PR.

PR #2494, opened 2025-01-09
fix(chat): refinements to chat interface

nokome:
Current focus on improving general layout, multi-modal interactions, and insertion back into model.


nokome:
Notes on what I have done on this. I'm planning to move onto other things now. So leaving some todo items below for you to pickup please @mike-parkin.

I have rearranged UI elements associated with prompts and models. They are close, but not exactly the same, as the design discussed in Discord:

image

- The prompt select is not styled quite right (the name of the prompt is the same colour as its description) but there does not seem to be a way to do that (for single option selects) - it just grabs the textContent of the selected `<option>` (whereas for multiselect you can define `getTag`)

- The prompt preview pops up the whole screen which is okay except that it is not vertical scrollable, also needs some spacing and rounding of corners? @mike-parkin 

image

- I have put the model selector, next to the model parameter sliders. This is _alot_ easier to implement than trying to separate them, and in my opinion, is a better place to have them anyway

image

- If you add multiple files they wrap along the button and can be removed

image

I have implemented the ability to insert multiple nodes in Rust and created an prelim version of a frontend component to test that out. The selector shows the type and ids. I did this for debugging purposes but I think it could actually be very useful for usability to show the use the node types (using "removable tags" like we do for models and files in shots above).  It would show the user exactly what they are about to insert into their document, a whole chat message, a code block, or two paragraphs and a list - and give them the ability to remove one or more accidentally selected nodes.

The insert popup needs styling and the selection interaction will need a fair bit of work: it is hard to know how to select things like code blocks. @mike-parkin 

[Screencast from 2025-01-10 11-17-41.webm](https://github.com/user-attachments/assets/9fecfa8f-c24f-457a-bb61-c4b1daa6c2d6)

PR #2499, opened 2025-01-12
fix(lexical codec): support unordered lists

mduckles:
fixed how lists decoding in the lexical codec to differentiate between unordered and ordered.


nokome (approved):
Thanks for this, looks good!

PR #2501, opened 2025-01-13
Ghost publishing: image handling

nokome:
This adds a new crate `node-media` which factors out some functionality we had elsewhere for marshaling images in a document into an image directory with hash filenames.

It uses this crate to extract images from documents, upload them to Ghost, and rewrite the URLs of nodes in the document as needed.

@timClicks: I'll pass this over to you to finish off.

I've done some preliminary testing with this Markdown file

```md
A Base64 encoded image:

![](data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAoAAAAKCAYAAACNMs+9AAAAFUlEQVR42mNk+8/wn4EIwDiqkL4KAZCkFDN4mQTvAAAAAElFTkSuQmCC)

A file relative to this doc:

![](./image.png)
```

### DIFF
--- a/rust/publish-ghost/src/lib.rs
+++ b/rust/publish-ghost/src/lib.rs
@@ -14,9 +14,12 @@ use common::{
     serde_json,
     serde_with::skip_serializing_none,
     strum::Display,
+    tempfile,
+    tokio::fs::remove_dir_all,
     tracing,
 };
 use document::{
+    codecs,
     schema::{shortcuts::t, Node, PropertyValueOrString},
     CommandWait, DecodeOptions, Document, EncodeOptions, Format, LossesResponse,
 };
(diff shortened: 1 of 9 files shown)


timClicks:
One thing to fix with the latest push is that references to local files get replaced with Ghost's URLs when `stencilla ghost publish --pull` is invoked.

nokome:
> One thing to fix with the latest push is that references to local files get replaced with Ghost's URLs when `stencilla ghost publish --pull` is invoked.

Right. I think for now at least, we just "deal" with that by documenting that.

rust/publish-ghost/src/lib.rs
```diff
@@ -164,11 +167,52 @@ impl Cli {
...
+                            file_name.to_string()
+                        }
+                    })
+                })
+            },
+        );
+
         // Construct the POST payload
-        let payload = Payload::from_doc(resource_type, &doc, None).await?;
+        let payload: Payload = Payload::from_doc(resource_type, &doc, None).await?;
```
timClicks, line 215:
```suggestion
        let payload = Payload::from_doc(resource_type, &doc, None).await?;
```

rust/publish-ghost/src/lib.rs
```diff
@@ -536,17 +637,42 @@ impl Payload {
...
+        //     &media_dir,
+        //     |_old_url, file_name| {
+        //         // Construct a new URL for the media file
+        //         // TODO: make this what it should be
+        //         format!("images/{file_name}")
+        //     },
+        // );
+
+        // // TODO: Upload images to Ghost. This should check that files don't already
+        // // exist on the server
```
timClicks, line 663:
```suggestion
```

timClicks:
It's strange to me that the build is failing in CI, as it's building fine on my machine.

Issue #2504, opened 2025-01-14
bug(codec-lexical): Paragraphs within lists


jduckles:
### What happened?

clap generated documentation creates markdown blocks like:

```
- `--log-level <LOG_LEVEL>` — The minimum log level to output

  Default value: `info`

  Possible values: `trace`, `debug`, `info`, `warn`, `error`
```

In Ghost this gets collapsed to:
<img width="721" alt="image" src="https://github.com/user-attachments/assets/52e53fdc-4f71-459f-9459-fa9fb1d0034f" />

### Which product?

Stencila CLI

### Which version?

main

nokome:
I tested this to work out if the issue resides with the Markdown codec or the Lexical codec. I put that snippet in `temp.md` and then did

```sh
cargo run -p cli convert temp.md --to json
```

which gives the expected structure (separate paragraphs nexted under a list item:

```json
{
  "type": "Article",
  "content": [
    {
      "type": "List",
      "items": [
        {
          "type": "ListItem",
          "content": [
            {
              "type": "Paragraph",
              "content": [
                {
                  "type": "CodeInline",
                  "code": {
                    "string": "--log-level <LOG_LEVEL>"
                  }
                },
                {
                  "type": "Text",
                  "value": {
                    "string": " — The minimum log level to output"
                  }
                }
              ]
            },
            {
              "type": "Paragraph",
              "content": [
                {
                  "type": "Text",
                  "value": {
                    "string": "Default value: "
                  }
                },
```

and then

```sh
cargo run -p cli convert temp.md --to koenig --pretty
```

which gives lacks the nested paragraphs and treats everything as inine text

```json
{
  "root": {
    "type": "root",
    "children": [
      {
        "type": "list",
        "listType": "bullet",
        "start": 1,
        "children": [
          {
            "type": "listitem",
            "children": [
              {
                "type": "extended-text",
                "format": 16,
                "text": "--log-level <LOG_LEVEL>"
              },
              {
                "type": "extended-text",
                "format": 0,
                "text": " — The minimum log level to output"
              },
              {
                "type": "extended-text",
                "format": 0,
                "text": "Default value: "
              }
```

@mduckles can you fix this in the Lexical codec please.

mduckles:
looking at the lexical from a ghost site with it correctly formatted there seems to be a linebreak inline that we aren't handling 
```  "children": [
              {
                "detail": 0,
                "format": 0,
                "mode": "normal",
                "style": "",
[...]

PR #2505, opened 2025-01-15
fix(lexical codec):  fix decoding and endcoding of lists

mduckles:
as discussed in #2504 ghosts lexical list items contain `linebreak` blocks which were being lost in `listitems`. this PR should fix this problem


mduckles:
@nokome through some more testing we found that we aren't handling nested lists from lexical. this seems to indicate that ghosts lexical and koenig don't distinguish between blocks and in-lines. e.g. here
```
{
  "root": {
    "children": [
      {
        "children": [
          {
            "children": [
              {
                "children": [
                  {
                    "children": [
                      {
                        "detail": 0,
                        "format": 0,
                        "mode": "normal",
                        "style": "",
                        "text": "second indent list",
                        "type": "extended-text",
                        "version": 1
                      }
                    ],
                    "direction": "ltr",
                    "format": "",
                    "indent": 1,
                    "type": "listitem",
                    "version": 1,
                    "value": 1
                  }
                ],
                "direction": "ltr",
                "format": "",
                "indent": 0,
                "type": "list",
                "version": 1,
                "listType": "bullet",
                "start": 1,
                "tag": "ul"
              }
            ],
            "direction": "ltr",
            "format": "",
            "indent": 0,
            "type": "listitem",
            "version": 1,
            "value": 2
          }
        ],
        "direction": "ltr",
        "format": "",
        "indent": 0,
        "type": "list",
        "version": 1,
        "listType": "bullet",
        "start": 1,
        "tag": "ul"
      }
    ],
    "direction": "ltr",
    "format": "",
    "indent": 0,
    "type": "root",
    "version": 1
  }
}
```

 This is a problem in this case as we are handling list items children as in-lines but in this case they can be blocks. should I make a `lexical::InlineOrBlock` enum and handle it that way or is there a better solution?

nokome:
> should I make a `lexical::InlineOrBlock` enum and handle it that way or handle all `listitem` children as blocks?

If you can do it by treating all list item children as blocks I think that would be better as that is how we do it it in our schema.

Issue #2495, opened 2025-01-09
feat(cli): Publish to Zenodo


jduckles:
### Which component?

Stencila CLI

### What's your idea?

@nokome has mentioned a desire to support publication of works to Zenodo.

This issue serves to give an overview of how Zenodo's API and metadata work to inform the effort to add that feature.

Some useful resources at Zenodo:
- [Zenodo Sandbox](https://sandbox.zenodo.org/) - helpful for testing (and should probably be an option in CLI so users can test their submissions before they go live for real) 
- [Zenodo GitHub Integration](https://developers.zenodo.org/#github) - Can deposit to Zenodo based on a `.zenodo.json` file for every release event of a GitHub repository.
- [Zenodo API](https://developers.zenodo.org/#quickstart-upload)
- API Tokens - [Register](https://zenodo.org/signup) for a Zenodo account if you don’t already have one. Go to your [Applications](https://zenodo.org/account/settings/applications/), to [create a new token](https://zenodo.org/account/settings/applications/tokens/new/).
- [Entity Metadata](https://developers.zenodo.org/#entities) - Partial (required) elements summarized below. 

Zenodo works are known as "Depositions" in Zenodo parlance.  

A deposition to zenodo has the following required metadata (along with other [non-required metadata](https://developers.zenodo.org/#entities)): 

| Attribute                               | Required                                  | Description                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                           |
| --------------------------------------- | ----------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| `upload_type`<br>_string_               | Yes                                       | Controlled vocabulary:<br>\* `publication`: Publication<br>\* `poster`: Poster<br>\* `presentation`: Presentation<br>\* `dataset`: Dataset<br>\* `image`: Image<br>\* `video`: Video/Audio<br>\* `software`: Software<br>\* `lesson`: Lesson<br>\* `physicalobject`: Physical object<br>\* `other`: Other                                                                                                                                                                                                                                                                                                                                                             |
[...]

PR #2508, opened 2025-01-20
Publishing docs to zenodo #2495

timClicks:
Features

- [x] defaults to sandbox
- [x] compiled doc uploaded as its own file (SWB only so far)
- [x] author affiliations translated
- [x] DOI reservation
- [x] `--force` flag for immediate publication
- [x] end-user messages
- [x] more metadata from document pre-populated into zenodo

Bugs to resolve

- [x] <strike>duplicate DOIs are created if the document is re-submitted from stencila</strike> [we're not storing the DOI in the document ourselves now, so this is not relevant
- [x] Upload default format to same as input format (not SWB)

Things to check

- The command-line help is rendered as intended. I have made the choice to break options into groups so that options that relate to this command are not mixed together with options relating to logging, etc.


rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+/// Publish to Zenodo
+/// 
+/// Metadata for the deposition is extracted from the metadata within the
+/// document.
+/// 
+/// Once uploaded, you will prompted to review the deposition and publish from
+/// Zenodo's web interface. If you wish to skip the review process and publish
+/// immediately, then use the `--force` flag.
+/// 
+/// By default, Zenodo's testing server is used. This means that DOIs created
```
jduckles, line 56: s/testing/sandbox to use Zenodo's language for the server.

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+/// will not resolve. To upload to Zenodo's public-facing production server, use
+/// the `--zenodo` flag. To upload to a specific server instance, use
+/// `--server`.
+/// 
+/// Unless told otherwise, your deposit will be given the publication type "preprint".
```
jduckles, line 61: Add short message with guidance on how to change this from `preprint` to other alternatives.

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+    /// 
+    /// Enable the "deposit:write" scope to enable the `--force` flag.
+    #[arg(long, env = "STENCILA_ZENODO_TOKEN")]
+    #[arg(help_heading("Zenodo Settings"))]
+    #[arg(display_order(1))]
+    token: Option<String>,
+
+    // Server selection options
+
+    /// Publish to Zenodo's testing server
```
jduckles, line 83: s/testing/sandbox

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+    /// 
+    /// When omitted, Zenodo will use today's date.
+    #[arg(long)]
+    #[arg(help_heading("Deposition Metadata"), display_order(3))]
+    #[arg(value_parser = parse_date)]
+    publication_date: Option<schema::Date>,
+
+    /// Title to use for the deposit
+    /// 
+    /// Required when the information is not available within the document.
```
jduckles, line 136: Maybe a message as to how you specify it in the document:

> Required when the information is not available within the document as an `Article:title` schema object.

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+    #[arg(long)]
+    #[arg(help_heading("Deposition Metadata"), display_order(3))]
+    title: Option<String>,
+
+    /// Description to use within the deposition
+    /// 
+    /// Required when the information is not available within the document.
```
jduckles, line 143: Maybe a message as to how you specify it in the document:

> Required when the information is not available within the document as an `Article:description` schema object.

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+    #[arg(long)]
+    #[arg(help_heading("Deposition Metadata"), display_order(3))]
+    description: Option<String>,
+
+    ///  One of the publication types from Zenodo's controlled vocabulary.
```
jduckles, line 148: Consider enumerating all options for the user in the long version of the help text with (--help) - 

`[ annotationcollection, book, section, conferencepaper, datamanagementplan, article, patent, preprint, deliverable, milestone, proposal, report, softwaredocumentation, taxonomictreatment, technicalnote, thesis, workingpaper, other ] `

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+    /// 
+    /// [default: preprint]
+    #[arg(long)]
+    #[arg(required_if_eq("publication", "true"))]
+    #[arg(default_value_if("publication", "true", "preprint"))]
+    #[arg(help_heading("Deposition Settings"), display_order(2))]
+    publication_type: Option<PublicationType>,
+
+    /// Force publish the deposition immediately
```
jduckles, line 157: add:

(use with care)

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+            let data: Value = deposition_response.json().await?;
+
+            if let Some(Value::Array(errors))  = data.get("errors") {
+                for error in errors {
+                    if let (Some(Value::String(field)), Some(Value::Array(messages))) = (error.get("field"), error.get("messages")) {
+                        if field == "metadata.description" {
+                            for message in messages {
+                                message.as_str().map(|msg| {
+                                    if msg == "Field may not be null." {
+                                        tracing::info!("hint: Provide a description with the --description flag.");
```
jduckles, line 421: ... or `Article:description` document metadata

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+                ))
+                .query(&[("access_token", &token)])
+                .send()
+                .await?;
+
+            if !publish_response.status().is_success() {
+                bail!("Failed to publish deposition: {}", publish_response.text().await?);
+            }
+
+            tracing::info!("Created published deposition at {deposition_url}.");
```
jduckles, line 532: If the deposition url is at "sandbox.zenodo.org" let's send another info message that alerts the user that they've only pushed to the sandbox. Something like: 

> You have successfully published to the Zenodo sandbox, please check that this deposition is accurate at the Sandbox URL and use the `--zenodo` flag to resubmit to the production Zenodo server.

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+mod cli_tests {
+    use super::*;
+
+    #[test]
+    fn defaults_to_sandbox() {
+        let args = Cli::parse_from(&["publish-zenodo"]);
+        assert!(args.sandbox);
+    }
+
+    /// Tests that the publication type defaults to pre-print when --publication is set
```
jduckles, line 558: for the Zenodo vocab it should be `preprint` not `pre-print`

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+    #[test]
+    fn publication_type_defaults_to_preprint() {
+        let args = Cli::parse_from(&["publish-zenodo", "--publication", "some.smd"]);
+        println!("{args:#?}");
+        assert!(matches!(args.publication_type, Some(PublicationType::Preprint)));
+    }
+
+    /// Tests that the publication type defaults to pre-print when --publication is set
```
jduckles, line 566: for the Zenodo vocab it should be preprint not pre-print

jduckles:
Looks great! Reading through the code I've made a few comments, mostly around user messaging. 

I wasn't sure why we are searching the document text for a DOI string? It seems we'll likely find DOIs in a reference section that aren't the correct DOI a large percentage of the time.

rust/publish-zenodo/src/lib.rs
timClicks: This already occurs thanks to `clap`. If you run `cargo run --bin stencila -- publish zenodo --help` from a checked-out version of this branch, then you'll see that the rendered documentation includes this.

rust/publish-zenodo/src/lib.rs
timClicks: This will be useful, but it's not really specific enough to be actionable (yet).

We could add an example for YAML front matter in here. But that could potentially pollute the purpose of help string, which is to document the option's behaviour.

It would be really nice to point to some documentation. There is no documentation on the website or similar about what the schema is or how to add items to it. The instructions need to be specific for each input format.

Asking to wade through the `/docs` within the repo feels like a big task. And as it's just reference material, it's very difficult to interpret for beginners who are thinking "I just want to add a title to a document"

nokome: Yes we absolutely need to get more guides and tutorials published on this soon.

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+            tracing::info!("Created published deposition at {deposition_url}.");
+        } else {
+
+            let mut msg = format!("Successfully published document as a draft deposit to {deposition_url}");
+            if let Some(doi) = reserved_doi {
+                msg.push_str(&format!(" with the doi:{doi} pre-reserved"));
+            }
+            msg.push('.');
+            println!("{msg}");
+            println!("Review and publish manually, or use --force to publish directly");
```
timClicks, line 541: Use the cli-utils rather than println macro.

rust/publish-zenodo/src/lib.rs
```diff
@@ -0,0 +1,575 @@
...
+    #[arg(default_value_if("lesson", common::clap::builder::ArgPredicate::IsPresent, "false"))]
+    #[arg(help_heading("Deposition Settings"), display_order(2))]
+    publication: bool,
+
+    /// Reserve a DOI for the deposition
+    #[arg(long)]
+    #[arg(help_heading("Deposition Settings"), display_order(2))]
+    reserve_doi: bool,
+
+    /// Publication date
```
jduckles, line 126: It is unclear what date format to use. An example here would be helpful.

rust/publish-zenodo/src/lib.rs
timClicks: This should now be addressed by the long help text which is provided to people when they invoke `stencila help publish zenodo` or `stencila publish zenodo --help`.

rust/publish-zenodo/src/lib.rs
timClicks: Help output now renders as 

```
      --publication-date <YYYY-MM-DD>
          Publication date
          
          Provide the date formatted as YYYY-MM-DD, e.g. 2012-03-10.
          
          When omitted, Zenodo will use today's date.
```

rust/publish-zenodo/src/lib.rs
timClicks: Fixed.

rust/publish-zenodo/src/lib.rs
timClicks: I've updated the help text.

Short form:

    --force                          Publish the deposition immediately (use with care)

Long form:

```plain
--force
    Publish the deposition immediately
        
    Requires that access token provided by the `--token` option has the "deposit:write" scope.
        
    WARNING: This is permanent. It will be impossible to review the deposition or make changes
    to it before it is publicly viewable.
```

rust/publish-zenodo/src/lib.rs
timClicks: Here is what the output looks like now from the command-line by default:

    $ export STENCILA_ZENODO_TOKEN="<TOKEN>" # from https://sandbox.zenodo.org/
    $ stencila publish zenodo <DOCUMENT_PATH>
    🎉 Draft deposition submitted
    🌐 URL: https://sandbox.zenodo.org/deposit/<deposit-id> (visit to check details and publish)
    📑 DOI: 10.5282/zenodo.<deposit-id>
    Note: This deposit has been submitted to the Zenodo Sandbox.
    Note: Use the --zenodo flag to resubmit to the production Zenodo server.

And this is the output when the user provides the `--zenodo` flag:

    $ export STENCILA_ZENODO_TOKEN="<TOKEN>" # from https://zenodo.org/
    $ stencila publish zenodo --zenodo <DOCUMENT_PATH>
    🎉 Draft deposition submitted
    🌐 URL: https://zenodo.org/deposit/<deposit-id> (visit to check details and publish)
    📑 DOI: 10.5281/zenodo.<deposit-id>

rust/publish-zenodo/src/lib.rs
timClicks: Corrected.

timClicks:
> I wasn't sure why we are searching the document text for a DOI string? It seems we'll likely find DOIs in a reference section that aren't the correct DOI a large percentage of the time.

This was some muddled Rust on my part. We only search the identifiers. When there are none present, we return early.

```rust
let Some(ids) = &article.options.identifiers else {
    return None;
};

for id in ids {
    if let Some(doi) = metadata_extraction::extract_doi(id) {
        return Some(doi.to_string());
    }
}
```

I've replaced the code with this block, which is more readable (hopefully):

```rust
if let Some(ids) = &article.options.identifiers {
    for id in ids {
        if let Some(doi) = metadata_extraction::extract_doi(id) {
            return Some(doi.to_string());
        }
    }
}
```

I tend to use the first style because I'm wary of rightward drift in Rust code.

PR #2518, opened 2025-01-22
feat(models): expose quality, cost, and speed scores

nokome:
@jduckles had the excellent idea of exposing model scores for overall quality (not specific to a particular prompt), cost, and speed in the select menu for models.

This implements that (currently with random scores pending adding those to `GET htpps://api.stencila.cloud/v1/models`) and more:

- the list of models is now sorted, across providers, in Rust, according to (a) routers always first, (b) quality score if any, (c) provider+model name

- the scores are displayed in the `stencila models list` command as colour coded numbers (and also in JSON or YAML when using the `--as` option)

Screenshot from 2025-01-23 08-04-34

- improved how models names & versions are displayed in the model selection menu which makes it consistent with the models sidebar menu

- the scores are displayed in the model selection menu using star, dollar and zap icons, examples below for wide and narrow displays

Screenshot from 2025-01-23 08-17-11

Screenshot from 2025-01-23 08-18-34

- tool tips in the menu like "Overall model score: 82/100" 

The information density is high (i.e. its pretty busy!) but I think this is a really good way to bring our evals/benchmarking right into the application and make it useful to users. Because there are correlations between quality and cost and speed - I think it will look less busy with real scores.

### DIFF
--- a/rust/model/src/lib.rs
+++ b/rust/model/src/lib.rs
@@ -83,6 +83,9 @@ pub struct ModelSpecification {
     version: String,
     r#type: ModelType,
     availability: ModelAvailability,
+    quality_score: Option<u32>,
+    cost_score: Option<u32>,
+    speed_score: Option<u32>,
 }
 
 impl From<&dyn Model> for ModelSpecification {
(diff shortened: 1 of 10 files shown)


jduckles:
It is busy, but I think its a good step to help users navigate things. 

This may not be possible, but it feels like you could get more information density with a sparkline bargraph for each icon. With three different colors for the three values (cost, quality, speed). It may not read quite as well visually though. 🤔

There are sparkline fonts out there -> https://github.com/aftertheflood/sparks?tab=readme-ov-file

image

jduckles (approved):
Looks great for a first go at aiding users in navigating an ever-growing list of models.

Issue #2533, opened 2025-01-28
feat(cli): `stencila publish ghost` add yaml flags


jduckles:
### Which component?

Stencila CLI

### What's your idea?

Some of the Ghost CLI metadata should come from document metadata itself: 

- title
- description: (rename from exceprt and rename flag `--excerpt` to `--description`) 

There are a few of the CLI flags for this subcommand that would be nice to have available in yaml headers. Since these aren't exactly document metadata I suggest we give them a `ghost-*` name by convention:

- `ghost-type`: [page / post] (equivalent to specifying `--page` or `--post`
- `ghost-slug`: slug (equivalent to specifying `--slug`) 
- `ghost-featured`: true / false ( true equivalent to specifying `--featured`
- `ghost-state`: published / draft (equivalent to `--draft` `--publish`)
- `ghost-schedule`: ISODATE (equivalent to `--schedule`)
- `ghost-tags`: allow yaml list or single string, augmets `--tag` functionality. 

This issue then can be closed when the following are done:
- [ ] Rename `--excerpt` to `--description` and pass `description:` metadata to that value
- [ ] Allow for `ghost-type` yaml metadata
- [ ] Allow for `ghost-slug` yaml metadata
- [ ] Allow for `ghost-featured` yaml metadata
- [ ] Allow for `ghost-state` yaml metadata
- [ ] Allow for `ghost-schedule` yaml metadata
- [ ] Allow for `ghost-tags` yaml metadata

jduckles:
Based on discussions this morning (January 29, 2025) we're going to change tacks a bit.

We'll have a `publish` key that publish subcommands use, the sub command `ghost` will have its own sub-key of `publish`. See example below. 

```
---
title: My research report 
description: A research report on something very interesting
access-right: open
authors: ["John Doe", "Jane Doe"]
publish: 
  ghost:
    type: page / post
    slug: sort-name-for-post
    featured: true
    state: draft # or publish
    schedule: ISODATE`
    tags: ["Documentation", "#docs"]
---

Note the `publish:` yaml key could also have have a `zenodo:` (or other yet-to-be-created publish targets) sub key which can be used to set many of the same values you can set with commandline flags. 

We should find a compact way to document these yaml configuration parameters too.

**Comment by jduckles** [comment]
@mduckles once you've added this to ghost, there will be a few options to add to the zenodo subcommand as well:

```
---
title: My research report 
description: A research report on something very interesting
access-right: open
[...]

PR #2540, opened 2025-01-30
publish ghost and zenodo yaml config header

mduckles:
I have implemented what was discussed in  #2533 for ghost

While testing it having `schedule` be separate to `draft` and `publish` can lead to problems as they are mutually exclusive when sent to ghost.
``` yaml
config:
  publish:
    ghost:
    state: draft # or publish
    schedule: ISODATE
```
as this is clear whether it should be a draft or scheduled.


jduckles:
Hey @nokome I talked to @mduckles and he said this is good to go now. Would be useful to have it merged for adding documentation metadata.

PR #2550, opened 2025-02-03
Fix/codemirror keybindings

mike-parkin:
**details**

- makes sure the editable codemirror views clipboard events cut and paste fires the patch-node cmd
- attaches a 'ctrl-Enter' keybinding to run the node, have to override the default here.
- Also makes the inlcude block source property editiable in the ui, with the smae ctrl-enter execute command

The `patchValueExecute` command kept failing in vscode lsp, but the command was being dispatched from the webview successfully.


nokome:
Thanks Mike. I have fixed issues and will merge.

PR #2551, opened 2025-02-03
refactor(web): move all node-related styles into seperate file

mike-parkin:
**details**

refactors the styles for stencila 'nodes' out of the default.css theme file, this allow the rules for the nodes to be used independently, without having to be wrapped in a view or [root] node.

### DIFF
--- a/web/src/themes/default.css
+++ b/web/src/themes/default.css
@@ -5,21 +5,27 @@
 @import url('https://fonts.googleapis.com/css2?family=Source+Serif+4:ital,opsz,wght@0,8..60,200..900;1,8..60,200..900&display=swap');
 @import url('../fonts/stix-two-math/stix-two-math.css');
 
+@import './nodes.css';
+
 :root {
   /* custom properties */
   --default-text-colour: #171817;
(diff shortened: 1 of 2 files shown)


web/src/themes/default.css
```diff
@@ -391,15 +79,9 @@ body {
...
-    }
-
-    stencila-datatable {
-      table {
-        @apply mx-0;
-      }
-    }
+    } */
 
     stencila-figure,
```
nokome, line 86: Can these also be moved to node.css? Do they still need to be nested under `[view='dynamic'], [view='vscode']`?

nokome (approved):
Thanks Mike. Left one comment. Happy to merge if you are confident it doesn't break things. I haven't checkd Argos yet.

web/src/themes/default.css
nokome: I meant `stencila-figure` and all the `stencila-*` styles below it.

mike-parkin: I'm not acutaly sure why these are here, so I've just left them there, as they are different from the styles that are already applied to the nodes

mike-parkin:
will try to match up these argos results results first, looks like the page horizontal padding has changed

PR #2555, opened 2025-02-06
fix(web): stop text overflow in author elements

mike-parkin:
**details**

enforce the overflow ellipsis on the author node `<stencila-ui-node-author>` to stop details overflowing container


mike-parkin:
Have increased the width, from 72 to 75ch, and removed all additional rules

PR #2584, opened 2025-03-05
fix(vscode): add loading screen to preview panel

mike-parkin:
**details**

adds a loader screen to the webview panel for the preview

the loader is taken down when the window onload event is fired


nokome:
Thanks for this. The problem with adding a spinner component was that it was not activated until our JS bundle finished loading. So I took the approach of putting the spinner into the webview HTML as an SVG and loading our JS with `async`.

PR #2586, opened 2025-03-05
Fix/add mermaid bundler

mike-parkin:
**details**
there seems to  compatibility issues between parcel and mermaid ^11.4.

mermaid v11 has added two peer deps, @mermaid-js/parser and 'langium' in order to work, and parcel is having issues resolving the imports, if aliases are added to the package.json to account for these it still runs into problems with the dynamically imported chunks. After testing here vite does not have the same problems so can compile 

As one solution I have created a separate package that will bundle the mermaid package with vite, and be imported by the web packages as '@stencila/mermaid'

**this does add some more vulnerabilities to the npm audit, most of which seem to come from vite**


nokome:
Thanks Mike. I am hesitant to add another module and another bundler as a dev dependency in this repo (we already have Parcel, Rollup and ESBuild). Just adds more stuff to maintain. Let me sit on it for a while.

PR #2725, opened 2026-04-15
local ollama support

stevejpurves:
a naive vibe coded experiment to add ollama, works for model selection when overide added to `toml` file but when in session, slow and returns json to tui, or unexpected tui controls, etc...


nokome:
Thanks for this Steve! I've had a quick skim and it looks good but I'll need more time to assess whether some of the architectural choices are in the right direction.

stevejpurves:
Totally, I am seeing a big difference in how the llama model, interacts with the agent definitions and tools which also might need some figuring out, going to try different models also. Meant this as a spike.

